PR #7, opened 2016-08-03
create the streaming endpoint

rybit:
this should create the ability to hook up to `streaming.netlify.com/api/v1/streaming` and
then handle messages coming from that endpoint


rybit:
nm not ready

PR #22, opened 2017-04-16
Remove duplicated `review_id` property.

pdhodgkinson:
There was a duplicated _review_id_ property in the _deploy_ definition.
This causes a problem with some tools that rely on successful parsing of the swagger file. For example [JS-YAML](http://nodeca.github.io/js-yaml/)


calavera:
Thanks!

PR #24, opened 2017-08-29
Make sure we don't load all files in memory during a deploy

biilmann:
This is a quick fix to make sure we don't load all files into memory but stream them from disk instead while uploading.


calavera:
LGTM

PR #41, opened 2017-11-29
Use new swagger to re-generate code

brycekahle:
Notably adds timeout, context, and http client to all operation params


calavera:
LGTM

PR #45, opened 2017-12-01
Check for invalid characters instead of escaping

brycekahle:
This matches how the API handles these characters.


go/porcelain/deploy.go
```diff
@@ -399,7 +404,7 @@ func (n *Netlify) uploadFile(ctx context.Context, d *models.Deploy, f *FileBundl
 
 		switch t {
 		case fileUpload:
-			name := (&url.URL{Path: f.Name}).EscapedPath()
+			name := f.Name
```
rybit: this var seems unneeded now

PR #61, opened 2017-12-19
Complete site's build settings.

### LINKED ISSUE #57: SiteBuildSettings missing many key fields
Howdy! 👋 

`SiteBuildSettings` is missing a lot of really important fields I wanted to point out:

  * `env`
  * `provider`
  * `repo_type`
  * `repo_url`
  * `repo_branch`

I'm able to _create_ a site connected to a VCS from the API but I can't reload that data and verify. This is particularly important in Terraform so it can reconcile if anything changed remotely.

**EDIT:** Added `env` as an important one since I can't modify env vars until this is available.

calavera:
The build settings is basically a repository into object underneath.
This changes the definition to reuse the RepoInfo struct. `repo_branch`
is used a read/write attribute to set the production branch for a site.

@mitchellh `repo_type` is a deprecated field and we don't use it for anything. This change should address the rest of attributes.
I renamed `branch` to `repo_branch` because it's only one property that you can use it for read and write.

Fixes #57.


mitchellh:
Thanks! That's no problem and I noticed that there was some weird naming thing going on there but had no idea which way was the right way. I'll update this to work no problem. Thank you.

mitchellh:
Hey @calavera I noticed the frontend sends `branch`, and you changed the API to use `repo_branch` as you said. I haven't verified it either way but is this okay? They both work?

PR #66, opened 2017-12-20
Unify repository path accessors with RepoPath.

### LINKED ISSUE #64: Site.BuildSettings.Repo is not populated on GET
Hello,

For `CreateSite`, the API allows us to populate the `repo` field with something like `mitchellh/foo`. However, upon reading that site, only the `repo_url` is populated and `repo` is empty.

For Terraform, this is making it difficult to verify the remote settings are setup correctly. If this is working as intended I can try to parse this information out but the best thing would be to avoid that if possible.

calavera:
POST and GET use the same attribute to access this information.

Fixes #64
/cc @mitchellh

### DIFF
--- a/go/models/repo_info.go
+++ b/go/models/repo_info.go
@@ -44,11 +44,11 @@ type RepoInfo struct {
 	// public repo
 	PublicRepo bool `json:"public_repo,omitempty"`
 
-	// repo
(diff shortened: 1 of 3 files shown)


mitchellh:
Brilliant. This works great. Thanks for being so fast 🐎

PR #74, opened 2018-01-09
Do not keep all file handles open at the same time

### LINKED ISSUE #73: Deploy can open too many files at once
A file is opened to compute the SHA, but keeps the file handle open. If the deploy has a large number of files, this can result in a `too many open files` error.

brycekahle:
Fixes #73

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -7,7 +7,6 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
-	"hash"
 	"io"
 	"io/ioutil"
 	"os"
@@ -78,37 +77,23 @@ type uploadError struct {
 
 type FileBundle struct {
 	Name string
-	SHA  hash.Hash
+	Sum  string
 
+	// Path OR Buffer should be populated
+	Path   string
 	Buffer io.ReadSeeker
-}
-
-func (f *FileBundle) Sum() string {
-	return hex.EncodeToString(f.SHA.Sum(nil))
 }
 
 func (f *FileBundle) Read(p []byte) (n int, err error) {
 	return f.Buffer.Read(p)
 }
 
+func (f *FileBundle) Seek(offset int64, whence int) (int64, error) {
+	return f.Buffer.Seek(offset, whence)
+}
+
 func (f *FileBundle) Close() error {
 	return nil
-}
-
-// We're mocking up a closer, to make sure the underlying file handle
-// doesn't get closed during an upload, but can be rewinded for retries
-// This method closes the file handle for real.
-func (f *FileBundle) CloseForReal() error {
-	closer, ok := f.Buffer.(io.Closer)
-	if ok {
-		return closer.Close()
-	}
-	return nil
-}
-
-func (f *FileBundle) Rewind() error {
-	_, err := f.Buffer.Seek(0, 0)
-	return err
 }
 
 type deployFiles struct {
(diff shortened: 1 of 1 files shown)


calavera:
LGTM

go/porcelain/deploy.go
```diff
@@ -78,39 +77,25 @@ type uploadError struct {
...
-}
-
 func (f *FileBundle) Read(p []byte) (n int, err error) {
 	return f.Buffer.Read(p)
 }
 
-func (f *FileBundle) Close() error {
-	return nil
+func (f *FileBundle) Seek(offset int64, whence int) (int64, error) {
+	return f.Buffer.Seek(offset, whence)
```
rybit: doesn't this have the problem that the Buffer could not be populated?

go/porcelain/deploy.go
```diff
@@ -404,21 +387,27 @@ func (n *Netlify) uploadFile(ctx context.Context, d *models.Deploy, f *FileBundl
 
 		switch t {
 		case fileUpload:
-			params := operations.NewUploadDeployFileParams().WithDeployID(d.ID).WithPath(f.Name).WithFileBody(f)
-			if timeout != 0 {
-				params.SetTimeout(timeout)
+			body, operationError := os.Open(f.Path)
```
rybit: doesn't this mask the operationError? https://play.golang.org/p/R2gPzwewg3u

go/porcelain/deploy.go
```diff
@@ -404,21 +387,27 @@ func (n *Netlify) uploadFile(ctx context.Context, d *models.Deploy, f *FileBundl
...
 				params.SetTimeout(timeout)
 			}
 			_, operationError = n.Operations.UploadDeployFunction(params, authInfo)
+			if operationError != nil {
+				f.Buffer.Seek(0, 0)
+			}
 		}
 
 		if operationError != nil {
-			f.Rewind()
```
rybit: don't we want to open the file in both cases? Function of file upload? Then rewind in both cases?

go/porcelain/deploy.go
```diff
@@ -78,39 +77,25 @@ type uploadError struct {
 
 type FileBundle struct {
 	Name string
-	SHA  hash.Hash
+	Sum  string
 
+	// Path OR Buffer should be populated
+	Path   string
 	Buffer io.ReadSeeker
```
rybit: might just not be easy to find in this PR - but in looking at the walk function, don't we not set the Buffer any more? Meaning Seek will fail as well as the write for functions?

go/porcelain/deploy.go
brycekahle: Yes. I can add a check, but it would be programmer error if you attempted to use this interface without Buffer populated.

I could think about a safer way to do this, but it would be more changes.

go/porcelain/deploy.go
brycekahle: Yes, good catch.

go/porcelain/deploy.go
brycekahle: No. This is exactly what I'm fixing. If there are 10k files, we do not want to open all 10k at once and leave them open. For files (not functions), this PR computes the sums for all the files, and then opens them on demand and they are uploaded.

go/porcelain/deploy.go
brycekahle: It is set for function uploads only since we create a zip archive in-memory.

PR #76, opened 2018-01-18
Upload go binaries as functions

calavera:
With AWSLambda supporting Go, I thought it'd be interesting to
do some exploration to support them.

Go support requires binary files, so we cannot depend on the file
extension to know if what we're uploading is JS or Go. Instead, we can
have a well known path convention to put Go binaries, like
`well-known-go`, and assume that all executables there are Go binaries.

We'd also need a way to tell our API that something should use the go
runtime and not the js runtime. I added a `Runtime` attribute to the file
bundle, so we can configure aws lambda with the right flags.

### DIFF
--- a/go/Gopkg.lock
+++ b/go/Gopkg.lock
@@ -122,6 +122,12 @@
   version = "v1.0.0"
 
 [[projects]]
+  branch = "master"
+  name = "github.com/rsc/goversion"
+  packages = ["version"]
+  revision = "b792656d10497b6b61ef55ce8e3bb725965bda2c"
+
+[[projects]]
   name = "github.com/sirupsen/logrus"
   packages = ["."]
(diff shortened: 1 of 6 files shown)


rybit:
I don't love the 'well-known-go', but that is probably is just a naming thing. We could also try to check (1) is it executable (2) is it linux amd64 binary? I'm not sure if that is easy in code - I've not looked, but those are requirements for a successful binary. 

Also - we will have to do the bundling and such, which isn't bad at all (just zip it).

calavera:
> We could also try to check (1) is it executable (2) is it linux amd64 binary

It already checks if it's executable. I guess we can use `strings` or grep the binary to see if the header says that's a Go binary.

> Also - we will have to do the bundling and such, which isn't bad at all (just zip it).

We already zip it.

rybit:
can we check that it is a linux binary?

PR #81, opened 2018-04-07
Add forms functions to porcelain

lannonbr:
I added functions in the porcelain package to be able to list forms as well as list form submissions. This is a prerequisite to add functionality to list these inside netlifyctl.

### DIFF
--- a/go/plumbing/operations/list_forms_parameters.go
+++ b/go/plumbing/operations/list_forms_parameters.go
@@ -21,7 +21,7 @@ import (
 // NewListFormsParams creates a new ListFormsParams object
 // with the default values initialized.
 func NewListFormsParams() *ListFormsParams {
-
+	var ()
 	return &ListFormsParams{
 
 		timeout: cr.DefaultTimeout,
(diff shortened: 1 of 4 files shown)


calavera:
awesome! Thanks so much!

calavera:
I'm looking at the spec and I'm not sure if this is exactly correct. In this current configuration, listForms will list all forms for sites that belong to your authenticated user. You probably want only forms for a specific site. In order to get forms only for a specific site, we need to pass the "site_id" parameter in the query.

Do you mind adding that parameter to the swagger spec?
It should go here: https://github.com/netlify/open-api/blob/master/swagger.yml#L847
You can copy the parameter definition from this other endpoint, but removing `required: true`: https://github.com/netlify/open-api/blob/master/swagger.yml#L876

lannonbr:
Well, I currently am thinking of having it when I run `$ netlifyctl forms` to list a table of all forms with a few fields such as the form id and name, the site id/name, and the number of submissions to the form. I could create a ListFormsBySiteId function after implementing that site_id param as well so then a user could run `$ netlifyctl forms --site-id <site_id>" and get just the forms for a particular site. What would you recommend?

lannonbr:
There actually was a duplicate endpoint for /sites/{site_id}/deploys/{deploy_id} so I removed one of them. As well, I just added both the parameter and a ListFormsBySiteId function. after a few failing tests, everything should work now on the API side

lannonbr:
I found there is an endpoint of /sites/{site_id}/forms to get a site's forms which is in the spec here: https://github.com/netlify/open-api/blob/master/swagger.yml#L147. Should I revert the changes and use that endpoint, or just leave my code as is @calavera?

lannonbr:
@calavera any progress looking over this?

calavera:
LGTM

PR #87, opened 2018-05-23
Replace backslashes in file path names.

### LINKED ISSUE #86: update to allow paths to work on windows via command prompt - helps netlifyctl
https://github.com/netlify/netlifyctl/issues/109 has the details - and is blocked on this fix.

calavera:
Fixes issues not deploying files under the right paths using Windows.

Fixes #86

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -460,10 +460,11 @@ func walk(dir string, observer DeployObserver) (*deployFiles, error) {
 		}
 
 		if !info.IsDir() && info.Mode().IsRegular() {
-			rel, err := filepath.Rel(dir, path)
+			osRel, err := filepath.Rel(dir, path)
 			if err != nil {
 				return err
 			}
+			rel := forceSlashSeparators(osRel)
 
 			if ignoreFile(rel) {
 				return nil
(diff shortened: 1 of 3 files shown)


go/porcelain/deploy.go
```diff
@@ -460,10 +460,11 @@ func walk(dir string, observer DeployObserver) (*deployFiles, error) {
 		}
 
 		if !info.IsDir() && info.Mode().IsRegular() {
-			rel, err := filepath.Rel(dir, path)
+			osRel, err := filepath.Rel(dir, path)
 			if err != nil {
 				return err
 			}
+			rel := cleanFileName(osRel)
```
rybit: you could use `filepath.Clean` https://golang.org/pkg/path/filepath/#Clean

calavera: I'm going to change the function name, it's misleading.

rybit: @brycekahle 
```
Finally, any occurrences of slash are replaced by Separator.
```

Isn't that what we want?

brycekahle: No, because on windows, separator = `\`

rybit: That means that for any windows machine, this will force `/` not `\`? If you guys say that is right then ok! You've got more context than me.

brycekahle: This is precisely the bug we are trying to fix! https://github.com/netlify/netlifyctl/issues/109

PR #89, opened 2018-06-02
Added account/sites POST endpoint

futuregerald:


calavera:
@futuregerald did you use a different go-swagger version to generate the Go files? there are a lot of unrelated updates.

futuregerald:
That's possible. I mentioned the messages that I got to you last week, I believe that I ran `go get...` after that which would mean the latest versions were installed. =\

futuregerald:
@calavera How would you like me to proceed here? I think the updated version was in the last PR I put in that was already merged.

PR #99, opened 2018-08-24
Add feature to create file.Sum with original sha included for asset mgmt

keiko713:
**This is the code for asset management.**

If `asset_management` is true inside of deploy object's site capabilities (`deploy.SiteCapabilities.AssetManagement`), we do an extra process during the `walk`, which is walk through all the files of the deploy and generate the sha (so that you can pass it to API).

The sha (`file.Sum`) is usually `sha1`, but the new version of sha will be `sha1:original_file_sha256`, **if** the file is the git-lfs pointer file.


swagger.yml
```diff
@@ -1566,6 +1566,11 @@ definitions:
           type: boolean
         review_url:
           type: string
+        site_capabilities:
```
keiko713: this is something I added to api side. serialized deploy object now has `site_capabilities`, which has several information regarding the site capabilities.
since this `site_capabilities` is not `key:value=string:string` hash (sometime value can be object), I... wasn't sure how to define this in swagger.
`asset_management` key inside of `site_capabilities` is only present (and be true) if asset management feature is on. otherwise, there will be no such key. I'm not sure if this definition is okay for that case also...
In the code, I'm also using `deploy.SiteCapabilities.AssetManagement` but I'm not sure how it's gonna be if there is no `asset_management` key.

PR #103, opened 2018-09-06
Regenerate go client using go-swagger 0.16.0

keiko713:
What title says. This generates some user operations, since it was added recently to swagger.yml but regeneration didn't happen at that time.

There are many changes but looks like they're mostly:

* removing all unnecessary code inside of `Validate`
* add a model for user
* use the option `--with-flatten=full` since it was the "default" with 0.12.0 (previous version), this allows upgrading the version to cause minimal changes (also test passes)
* travis uses 1.10 and 1.11 now

Overall this looks pretty safe to me, but I'm... not really familiar with this, so any advice for the way to test it out or proving that this is okay is welcome!


Makefile
```diff
@@ -16,7 +16,7 @@ deps: ## Download dependencies.
 	go get -u github.com/golang/dep/cmd/dep && cd ./go && dep ensure
 
 generate: validate ## Generate the API Go client and the JSON document for the UI.
-	swagger generate client -A netlify -f swagger.yml -t go -c plumbing
+	swagger generate client -A netlify -f swagger.yml -t go -c plumbing --with-flatten=full
```
bcomnes: This has to do with the flattening of the different models.

PR #106, opened 2018-09-11
Add feature to create file.Sum with original sha included for asset mgmt (again)

keiko713:
We tried to merge the PR https://github.com/netlify/open-api/pull/99, but the base was using `regenerate-go-client` branch when we click the merge button, which was the branch already merged into the master, so we kinda merged into the branch that "doesn't really exist" and didn't merge this into the master.
Let's merge it to the master for real this time.


bcomnes:
Werps.

PR #110, opened 2018-09-15
First draft of addon routes

bcomnes:
We needed these routes in the CLI, so this is to port them over to our open-api spec.  

Closes https://github.com/netlify/cli/issues/72


bcomnes:
I don't know what I'm doing so this isn't going to pass right away.

calavera:
@futuregerald added these endpoints in ##137

Issue #113, opened 2018-10-02
appropriately handle unicode filenames in site deploys


fool:
Situation: Enterprise customer Global Citizen Year ($12k contract -  https://www.dropbox.com/s/sa2xz75t4k1d8c0/-Signed%20Global%20Citizen%20Year.pdf?dl=0) is unable to successfully deploy their site from git.  While there is another problem also blocking, this one is 1/2 of the problem.  They have unicode filenames and their deploys error like this:

`failed during stage 'deploying site': Failed to execute deploy: Upload cancelled: updates/on-walking-sharing-space-and-our-mother%c2%89s-garden/index.html`

(from https://app.netlify.com/sites/wizardly-wing-6da16b/deploys/5bac50cd73f2cf0ea39d04ca)

you can see the problem more clearly in humio:

`Failed to upload file updates/on-walking-sharing-space-and-our-mother%c2%89s-garden/index.html" build_id=5bac50cd73f2cf0ea39d04cb context=production deploy_id=5bac50cd73f2cf0ea39d04ca error="[PUT /deploys/{deploy_id}/files/{path}][422] uploadDeployFile default  u0026{Code:422 Message:Missing entry in deploy for /updates/on-walking-sharing-space-and-our-motheru0089s-garden/index.html}`

(from https://cloud.humio.com/netlify-production/search?widgetType=list-view&query=5bac50cd73f2cf0ea39d04ca%20on-walking-sharing-space&live=false&start=1537945200000&end=1538031600000&fullscreen=false)

@bcomnes tracked this down to this repo and behavior:

When you make the API call to create a deploy for files that contain unicode characters in filenames, filenames in the initial manifest need to be unescaped (== actual names), but when uploaded (PUT to the API) must be escaped since the filename is in a URL

fool:
cc @phae guess we need to discuss this during our meeting - nobody seems to have the bandwidth to make the save for $12k

futuregerald:
i'm unable to test this at the moment:

        make all
        swagger validate swagger.yml
        2018/10/11 13:35:56
        The swagger spec at "swagger.yml" is valid against swagger specification 2.0
        swagger generate client -A netlify -f swagger.yml -t go -c plumbing --with-flatten=full
        unknown flag `with-flatten'
        make: *** [generate] Error 1

[...]

PR #114, opened 2018-10-03
Escape file paths before using them in URL path

bcomnes:
Closes https://github.com/netlify/open-api/issues/113

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/go-openapi/errors"
 	"github.com/rsc/goversion/version"
+	"net/url"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


bcomnes:
Still WIP because I'm not 100% sure this fixes the issue yet.

bcomnes:
Might need http://golang.org/pkg/net/url/#PathEscape instead.  I'll investigate tomorrow.

fool:
cc @rybit could you spend 20 minutes today reviewing / attempting to fix up if needed?

bcomnes:
This is purely an idea based on an issue I ran into with the JS client, I never got a chance to experiment if this works or not.

bcomnes:
We haven't seen this in ages, so closing for now.

PR #116, opened 2018-10-12
Add getAccount route

bcomnes:
Add the getAccount operation.


bcomnes:
Includes a go client regeneration.

PR #126, opened 2019-01-07
Read original size from lfs pointers.

calavera:
So we can store that information.


keiko713:
What are you trying to do with this info in api side? I see "So we have a double guarantee about the content" in the second commit comment, maybe you want to do some checks...?
A bit more explanation would be helpful!

PR #136, opened 2019-02-01
added service instance endpoints and service manifest endpoint

futuregerald:
Added endpoints used in https://github.com/netlify/js-client/blob/master/src/addons.js and https://api.netlify.com/api/v1/services/${addonName}/manifest


swagger.yml
```diff
@@ -19,7 +19,7 @@ responses:
   error:
     description: error
     schema:
-      $ref: "#/definitions/error"
+      $ref: '#/definitions/error'
```
bcomnes: I'm not opposed to `s/"/'` but lets do it in a separate PR.  Can you unstage these quotation changes?

futuregerald: We're using single quotes everywhere else, are you saying we should leave it as double quotes in this one location?

futuregerald: Done

bcomnes: I'm saying if we fix it, let's do it in a separate PR.

bcomnes: Whitespace/formatting changes are best chunked into a single commit.

futuregerald: Makes sense. Thanks. I fixed it so let me know if it's good to merge.

PR #146, opened 2019-02-25
Ignore nil deploys.

calavera:
Allow to pass options when there is no previous deploy.


keiko713:
Can I get more context...?  Is this for deploying the site/repo without Git provider?

calavera:
> Can I get more context...? Is this for deploying the site/repo without Git provider?

yeah, you can call this from a cli directly without creating a deploy first.

PR #147, opened 2019-02-26
Update go dependencies

bcomnes:
This should fix the missing consumer errors:

- https://github.com/go-openapi/runtime/issues/106
- https://github.com/go-openapi/runtime/commit/17da9adfdb3ac94433ae7ca9e974d8ee7dc73265

PR #149, opened 2019-02-28
Update go-swagger libraries to the latest version and update to go modules

### LINKED ISSUE #122: Handle 408 timeouts better in go client / text error decoder
```
res, err := b.apiClient.DeployFiles(absFilePath, absFuncPath, b.Branch, b.GitRef(), b.SysLog)
```

When the retry uploads fail with a 408 you end up with an error like:

```
deploying site': Failed to execute deploy: no consumer: "text/html; charset=UTF-8"
```

Not very friendly.  We probably need to special case handle 408s since its not returning JSON messaging.

bcomnes:
This PR turned into a mess.  It does a few things:

## Fix the missing consumer errors

The go-swagger runtime crashed on plain/html content types.  This was causing problems in the buildbot, so the first step to address that was to update the go-swagger runtime libs to get the following fix:

- https://github.com/go-openapi/runtime/issues/106
- https://github.com/go-openapi/runtime/commit/17da9adfdb3ac94433ae7ca9e974d8ee7dc73265

Update go-swagger to `v0.18.0` from [`94927f8c9742791a9aa46c6c3965f0a37025f41d`](https://github.com/go-openapi/runtime/tree/94927f8c9742791a9aa46c6c3965f0a37025f41d)

## Fix the failing tests

After updating the go-swagger libraries, this test started failing:

https://github.com/netlify/open-api/blob/6aaaf6f356416f61c0075f254ed7be07acd157f2/go/porcelain/http/http_test.go#L146-L179

The error was the following:

```
Post http://127.0.0.1:41076/: http: ContentLength=14 with Body length 28
```

This comes from deep within the http request client, and happens when the body exceeds the content length of the req.ContentLength property.  

https://github.com/golang/go/blob/61170f85e62f1326d42c4dbd8aa17ab4a1305a87/src/net/http/transfer.go#L387-L389

Rolling back to `autorest@8.1.0` from `autorest@v9.6.0`  makes the test pass again, but we should really confirm retry's with a body work.  

Whats happening is:

- In the newer autorest versions, it calls `rr.req.GetBody()`

https://github.com/Azure/go-autorest/blob/1f7cd6cfe0adea687ad44a512dfe76140f804318/autorest/retriablerequest_1.8.go#L47-L53

- This doubles the size of the body in the `req` object for some reason.  I haven't dug into why that is yet.
- For comparison: here is the new implementation of GetBody we are stuck with for now that seems to double the size body's size: 

https://github.com/go-openapi/runtime/blob/9c75b2b632f58f9cb20e0767a724d5ebc37c2047/client/request.go#L209-L220

- Here is the old implementation of GetBody 

https://github.com/go-openapi/runtime/blob/94927f8c9742791a9aa46c6c3965f0a37025f41d/client/request.go#L249-L254

- For comparison here are the two other strategies: 

https://github.com/Azure/go-autorest/blob/1f7cd6cfe0adea687ad44a512dfe76140f804318/autorest/retriablerequest_1.7.go

- (this is the version we effectively rolled back too)

https://github.com/Azure/go-autorest/blob/1f7cd6cfe0adea687ad44a512dfe76140f804318/autorest/retriablerequest.go 

## Port over to go-modules

I simultaneously tried to move the go module over to its own repo https://github.com/netlify/go-client/tree/open-api but that is still being discussed: https://github.com/netlify/go-client/pull/17

I wanted to get that work out asap since the buildbot is already on go-modules, so I ported that work into this PR. 

## Add test to cover `no consumer` errors

Add a tests to verify the client doesn't error out when it receives an html or text response. 

Closes #122
Closes [netlify/buildbot#362](https://github.com/netlify/buildbot/issues/362)


bcomnes:
Same error as https://github.com/netlify/open-api/pull/147, but with the new toolchain.

bcomnes:
What's left now is to write some retriable tests with a body.

bcomnes:
Ok, end of the day here for me, for tomorrow, I think its the interaction between:

- GetBody implementation in go-swagger https://github.com/go-openapi/runtime/blob/9c75b2b632f58f9cb20e0767a724d5ebc37c2047/client/request.go#L209-L220 
- This part of the prepare() method in autorest: https://github.com/Azure/go-autorest/blob/1f7cd6cfe0adea687ad44a512dfe76140f804318/autorest/retriablerequest_1.8.go#L47-L54

bcomnes:
Adding a replay test with the pinned version of autorest seems to work, so thats promising.

bcomnes:
The docs changes here are still a wip.

bcomnes:
Added a test to confirm we no longer will get the no-consumer error when we get an unexpected content type.

PR #153, opened 2019-03-07
Change embedded DefaultSyncFileLimit to 500

bcomnes:
We are also setting this as a default on BitBalloon, but in case anyone uses this elsewhere, they will get the correct default.

### DIFF
--- a/go/porcelain/netlify_client.go
+++ b/go/porcelain/netlify_client.go
@@ -8,7 +8,7 @@ import (
 	"github.com/netlify/open-api/go/porcelain/http"
 )
 
-const DefaultSyncFileLimit = 7000
+const DefaultSyncFileLimit = 500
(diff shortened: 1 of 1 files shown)


go/porcelain/netlify_client.go
```diff
@@ -55,7 +55,9 @@ type Netlify struct {
 }
 
 func (n *Netlify) SetSyncFileLimit(limit int) {
-	n.syncFileLimit = limit
+	if limit > 0 {
+		n.syncFileLimit = limit
+	}
```
bcomnes: This way, 0 value will take the default value given the way the buildbot initializes this client: https://github.com/netlify/buildbot/blob/5a312d8ce9ade738407c76651271f2b27353a6c9/api/api.go#L71-L72

PR #154, opened 2019-03-07
Add observer hook for deploy polling start

bcomnes:
Adds a observation hook for right before we poll for a deploy.


bcomnes:
Good idea, but it was decided we no longer need to poll here.  Closing this, and removing polling code.

PR #156, opened 2019-03-08
Fix text error bug

bcomnes:
This updates dependencies and adds an error response test.  Down the road it can be adapted to cover the no consumer error, which is caused by a non-json response.


go/porcelain/deploy.go
```diff
@@ -337,13 +335,6 @@ func (n *Netlify) WaitUntilDeployReady(ctx context.Context, d *models.Deploy, ti
 	return n.waitForState(ctx, d, timeout, "prepared", "ready")
 }
 
-func (n *Netlify) WaitForDeployComplete(ctx context.Context, d *models.Deploy, timeout time.Duration) (*models.Deploy, error) {
-	if timeout <= 0 {
-		timeout = doneProcessingTimeout
-	}
-	return n.waitForState(ctx, d, timeout, "ready")
-}
-
```
bcomnes: Dead code

go/porcelain/deploy_test.go
```diff
@@ -65,3 +76,29 @@ func TestAddWithLargeMedia(t *testing.T) {
...
+	server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		rw.Header().Set("Content-Type", "application/json; charset=utf-8")
+		rw.WriteHeader(408)
+		rw.Write([]byte(`{ "foo": "bar", "message": "a message", "code": 408 }`))
+	}))
+	defer server.Close()
+
+	httpClient := http.DefaultClient
+	authInfo := runtime.ClientAuthInfoWriterFunc(func(r runtime.ClientRequest, _ strfmt.Registry) error {
+		_ = r.SetHeaderParam("User-Agent", "buildbot")
```
keiko713: kinda nitpicking but is this `_ =` part needed?

bcomnes: Good question.

bcomnes: Nope! Good catch.

Issue #108, opened 2018-09-12
Add Split Testing Endpoint


rajington:
Noticed there were issues for missing endpoints but didn't see one for `/sites/{site_id}/traffic_splits`, feel free to close if it's because it's still in beta!

calavera:
ENOTENOUGHTIMETOADDALLENDPOINTS :laughing: 

Feel free to add it, if you have time to inspect requests/responses. There is no real reason why it's not here.

PR #161, opened 2019-06-19
added split testing endpoints

futuregerald:
Added split testing endpoints to address https://github.com/netlify/open-api/issues/108


DavidWells:
Looks good to me.

Wish I knew about these a while ago 😄

PR #169, opened 2019-08-06
Remove `payment_method_id` from accountUpdateSetup

kitop:
It's not being processed by API.


mraerino:
@kitop is it still true that this field is not doing anything right now?

kitop:
> @kitop is it still true that this field is not doing anything right now?

Yup! Still true

mraerino:
Cool. I'm going to add a Regeneration of the go client to the branch and them merge

PR #176, opened 2019-09-09
Add `description` field

ehmicky:
This adds a [`description` field](https://swagger.io/specification/#infoObject) which shows up on top the documentation explaining what Netlify API does.


swagger.yml
```diff
@@ -3,6 +3,25 @@ info:
...
+  description: >-
+    Netlify is a hosting service for the programmable web. It understands your
+    documents and provides an API to handle atomic deploys of websites, manage
+    form submissions, inject JavaScript snippets, and much more. This is a
+    REST-style API that uses JSON for serialization and OAuth 2 for
+    authentication.
+
+
+    This document describes common usage in some detail. There is also an
+    OpenAPI reference for the Netlify API that you can explore. Visit our
```
verythorough: The "OpenAPI reference" mentioned here is this very page! I would reverse those two descriptions. In other words, explain that this site is a reference, and for more detailed instructions for common uses, visit netlify.com/docs/api.

swagger.yml
```diff
@@ -3,6 +3,25 @@ info:
...
+    Community forum to join the conversation about understanding and using
```
verythorough: I'd also add back the community link that's in that intro at netlify.com/docs/api.

ehmicky: Fixed in a5ad674022a55d0b4b3210f03cb5d4c4fac83ce0

swagger.yml
ehmicky: Fixed in f126c128ec640331844f497f6c9ad6967b631d84

ehmicky:
Thanks for the review @DavidWells and @verythorough! 
I fixed the description, ready for another round of review @verythorough.

ehmicky:
Since it does not work with Swagger UI, I reverted this commit on `master` (6c7bcbb83e9b2e6bbdd12b9477c1fb72a22f90d0) but it's added back by #182.

PR #181, opened 2019-09-09
Upgrade all dependencies

ehmicky:
This upgrades all dependencies.

I've checked all changelogs, and everything is ok.


package.json
```diff
@@ -8,27 +8,27 @@
...
+    "postcss-browser-reporter": "^0.6.0",
+    "postcss-cli": "^6.1.3",
+    "postcss-import": "^12.0.1",
+    "postcss-reporter": "^6.0.1",
     "postcss-url": "^8.0.0",
     "pump": "^3.0.0",
-    "rimraf": "^2.6.2",
-    "split2": "^3.1.0",
-    "swagger-ui": "^3.21.0"
+    "rimraf": "^3.0.0",
```
DavidWells: is `rimraf` 3x same API?

ehmicky: Yes it is. I also checked the other dependencies (in case they did not respect semver).

PR #182, opened 2019-09-09
Enhance documentation UI

### LINKED ISSUE #172: Enhance the HTML template
The current template is `swagger-ui` which is great. However there are alternative projects that might provide with a better documentation UX and UI such as [`widdershins`](https://mermade.github.io/shins/#swagger-petstore) and [`redoc`](https://redocly.github.io/redoc/).

What do you think?

ehmicky:
Fixes #172

This changes the HTML/CSS of the OpenAPI documentation using [`redoc`](https://github.com/Redocly/redoc) instead of [`swagger-ui`](https://github.com/swagger-api/swagger-ui).

## Before (screenshot)

before

## After (screenshot)

after

Before (URL): https://5d7672277da50b0008d42dda--open-api.netlify.com
After (URL): https://deploy-preview-182--open-api.netlify.com

Improvements over previous setup:
  - prettier
  - use Netlify branding (logo, colors, fonts)
  - mobile-friendlier
  - search bar
  - can add custom documentation sections, for example for Pagination, Rate limiting, etc. Those sections are not added yet.
  - possibility to show different code examples for Go, Node.js, HTTP or other languages/protocols. Note that our specification does not have code examples yet.1
  - much simpler setup. Redoc does everything: bundling, local server, etc. No need for us to maintain that code anymore.

Not changed:
  - support all OpenAPI keywords

To be done:
  - allow sending requests from the browser like in Swagger UI. I'm not sure that one is possible but I can look into it. We should probably recommend importing the OpenAPI definition in Postman or Insomnia instead anyway, as this provides with a better developer experience.
  - add sections for models (as opposed to endpoints)
  - lots of dependencies and code can be removed since `redoc` does everything. I can do the cleanup after this PR is merged.
  - add favicon


DavidWells:
This is super awesome 🎉

ehmicky:
Upgraded the logo (thanks @iKristy!).

DavidWells:
One note:

We will need to update `netlify api --list` command links when we take this live.

image

verythorough:
Oof, bummer redirects won't work since those are all behind a hash. :(

DavidWells:
We can handle redirects client side. 

`if hash && match('#/default/') > redirect to new url `

`#/default/createSite` to `#operation/createSite`

redoc/build.js
```diff
@@ -0,0 +1,87 @@
...
+    copyAssets(),
+  ])
+  // Inject analytics
+  await addAnalytics()
+}
+
+const redocCli = async function() {
+  await pExecFile('redoc-cli', [
+    `--title=${TITLE}`,
+    `--options.expandResponses=${SUCCESS_STATUS_CODES}`,
```
DavidWells: Thoughts on disabling this? 

When 200 response are auto expanded by default it makes scrolling much much longer.

I tried the build with this line commented out and it was easier to quickly scroll down the page.

ehmicky: Fixed in 275ecc4afc61484c1a8f3ffcf9a5fbb2a8a22125

ehmicky:
I've added the following:
  - favicons
  - `<meta>` tags (copied from `netlify-www`)
  - sections copied from the online documentation: "Making a request", "Authenticating", "Rate limiting", "Pagination", "Site IDs"

The endpoints are also now grouped by models. The models themselves are grouped when they are related.

When it comes to the links (that you mentioned above), I've opened https://github.com/netlify/cli/pull/546 to update the links of `netlify api --list`.

I have removed all the code related to the previous setup (based on Swagger UI).

Finally I made the following refactoring changes:
  - split into smaller files
  - moved all JavaScript-related code to `src/` and API documentation to `src/docs/`, so it's easier to distinguish form the Go-related code at `go/`
  - separate the injected JavaScript to its own `script.js` and the injected HTML to its own `head.html` so it's easier to add more to them

I need some feedback on the following:
  - I named the endpoint groups ("tags" and "tag groups") based on my intuition. However since I don't know the API I might have gotten them wrong. Could you please look at the left sidebar and make sure the grouping and the group names make sense?
  - Is there anything left to do when it comes to the links and redirects?
  - I would need an additional code review for this big PR, thanks @DavidWells and @verythorough! :)

ehmicky:
@verythorough fixed!

Is there anything left because this can be deployed to production?

ehmicky:
Oh I did not understand @DavidWells comment on the redirects, I get it now. Very good point!

I've fixed it in 60a7f7af4988f6811bbb6dcf6ed9017bfb4a4719

You try it for example https://deploy-preview-182--open-api.netlify.com/#/default/listSites

Ready for another round :)

PR #185, opened 2019-09-09
Add linting and prettifying

ehmicky:
This adds ESLint and Prettier.

The configuration is taken from `@netlify/cli`. 

Note that I changed some `'` to `"` in npm scripts so that this [works on Windows](https://github.com/ehmicky/cross-platform-node-guide/blob/master/docs/4_terminal/shell.md#cmdexe).


sw-yx:
we oughtta have this everywhere tbh.

PR #193, opened 2019-09-10
Fix YAML syntax error

ehmicky:
The current `splitTests` definition is parsed as:

```js
{ 
  type: 'array', 
  items: null, 
  '$ref': '#/definitions/splitTest' 
}
```

Instead of:

```js
{ 
  type: 'array', 
  items: {
    '$ref': '#/definitions/splitTest' 
  } 
}
```

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -1510,7 +1510,7 @@ definitions:
   splitTests:
     type: array
     items:
-    $ref: '#/definitions/splitTest'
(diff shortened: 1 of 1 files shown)


DavidWells:
Good eye!

ehmicky:
I actually caught it thanks to automated OpenAPI validation added by #194.

PR #194, opened 2019-09-10
Validate and normalize OpenAPI specification

ehmicky:
Validate the OpenAPI specification at build-time (using their official library [`swagger-parser`](https://github.com/swagger-api/swagger-parser)). This ensures we did not do enter any invalid OpenAPI keywords or syntax.

This also dereferences the JSON references (like `$ref: '#/responses/model'`) so that users of this library do not have to. This actually fixes some bugs. For example [here](https://github.com/netlify/js-client/blob/3d40ef4b1ba8523277bb25eb87398cd46d15ff17/src/open-api/index.js#L60) we assume the JSON references are dereferenced but they are not.

At the moment, two identical files are used to perform this build task. This is due to how Swagger UI is currently setup, but this issue should be fixed by #182.


ehmicky:
There are some merge conflicts to fix on this PR (`package-lock.json` removed, linting failing) that I am going to fix before merging this PR.

PR #170, opened 2019-08-13
Regenerate go client

Mongey:
So that #164 can be used

Issue #199, opened 2019-09-23
Tags are breaking the structure of the Go client


mraerino:
Tags on operations cause the Go client generator to create a package for every tag.
Although this seems reasonable we cannot use this feature because it would be a breaking change for all users of the Go client.

Current workaround is to remove tags before generating.
We need to think of a stable way to keep doing this or make a breaking change, e.g. v2 release.

ehmicky:
Should this be closed?

mraerino:
Let's discuss how to solve the rough edges of my workaround over here.

ehmicky:
Three questions: 
  - are we always sure the `Makefile` will be executed on a machine that has `sh` (e.g. some Windows machines don't) and `jq` installed?
  - we are already both validating and "flattening" the `swagger.yml` and [converting](https://github.com/netlify/open-api/blob/master/src/convert.js) it to `swagger.json` as part of [`npm run convert`](https://github.com/netlify/open-api/blob/da4c216d4d3dc41989467aef21f3345a8564f61d/package.json#L56). This could be used instead of `go-swagger` `validate` and `flatten` methods?
  - likewise we could remove the `tags` properties from that `convert.js` method, and output it to a second file `swagger_go.json`?

verythorough:
I don't suppose that would mean we could move away from go-swagger and upgrade to OpenAPI 3.0?

ehmicky:
While we could use the Node.js logic instead of `go-swagger` for validation, flattening and JSON conversion, we still need `go-swagger` to generate the Go SDK code.

mraerino:
> I don't suppose that would mean we could move away from go-swagger and upgrade to OpenAPI 3.0?

I'm happy to explore that. is there an existing issue?

ehmicky:
There is an issue at #171.

PR #200, opened 2019-09-23
Fix Go client generation

mraerino:
*Closes #170*

## Summary

*Updates to Go 1.13*

Uses a workaround to ignore `tags` when generating the Go client.
Generating the Go client while the operations have different tags attached means generating a different package for every tag. This breaks the API for all users of the client because they will most likely rely on types for request params even if they are just using the porcelain package.

I opened https://github.com/netlify/open-api/issues/199 to track this issue.


ehmicky:
Thanks a lot for this @mraerino!

Three questions: 
  - are we always sure the `Makefile` will be executed on a machine that has `sh` (e.g. some Windows machines don't) and `jq` installed?
  - we are already both validating and "flattening" the `swagger.yml` and [converting](https://github.com/netlify/open-api/blob/master/src/convert.js) it to `swagger.json` as part of [`npm run convert`](https://github.com/netlify/open-api/blob/da4c216d4d3dc41989467aef21f3345a8564f61d/package.json#L56). This could be used instead of `go-swagger` `validate` and `flatten` methods?
  - likewise we could remove the `tags` properties from that `convert.js` method, and output it to a second file `swagger_go.json`?

mraerino:
Those are valid remarks. Can we move this to #199?

PR #205, opened 2019-12-12
Fix options of go generation command

mraerino:
Without this option some files were structured differently when regenerating. So adding this for consistency.


keiko713:
seems like you're the one who removed it, maybe you accidentally removed it last time? if so, totally +1

https://github.com/netlify/open-api/pull/200/files#diff-7017bb2aedbbeb1e05b6fbffb0f46f41R6

mraerino:
Yeah, seems like I did accidentally remove it. 🙈

PR #209, opened 2020-01-14
only check deploy overcommited if no files are returned directly

ingride:
part of simplifying the logic on the client side, when we have a direct response with the required files

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -257,7 +257,7 @@ func (n *Netlify) DoDeploy(ctx context.Context, options *DeployOptions, deploy *
 		deploy = resp.Payload
 	}
 
-	if (len(deploy.Required) == 0) {
+	if len(deploy.Required) == 0 {
 		if n.overCommitted(options.files) {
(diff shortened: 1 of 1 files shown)


go/porcelain/deploy.go
```diff
@@ -257,14 +257,16 @@ func (n *Netlify) DoDeploy(ctx context.Context, options *DeployOptions, deploy *
 		deploy = resp.Payload
 	}
 
-	if n.overCommitted(options.files) {
-		var err error
-		deploy, err = n.WaitUntilDeployReady(ctx, deploy, options.PreProcessTimeout)
-		if err != nil {
-			if options.Observer != nil {
-				options.Observer.OnFailedDelta(deployFiles)
+	if (len(deploy.Required) == 0) {
```
smoya: Minor: unneeded extra parentheses

ingride: thanks, i caught it too, just forgot to push 🤦‍♀

ingride:
going in a different direction

PR #217, opened 2020-02-25
Document deploy cancel

vbrown608:
Documents the endpoint to cancel an in-progress build/depoy.


swagger.yml
```diff
@@ -628,6 +628,26 @@ paths:
             $ref: '#/definitions/deploy'
         default:
           $ref: '#/responses/error'
+  /sites/{site_id}/deploys/{deploy_id}/cancel:
```
kitop: Late to the party, but URL for this is `/deploys/{deploy_id}/cancel`, no need for the `site_id` prefix.

PR #222, opened 2020-03-11
Fix path for deploy cancel

vbrown608:
Fixing the path for canceling a deploy. Thanks for catching this @kitop !


ehmicky:
Thanks a lot @vbrown608 for fixing it and @kitop for pointing out at the solution!

PR #224, opened 2020-04-14
Warn if function conditions are not met

mraerino:
### Summary

I recently discovered that we do not deploy go functions with invalid permissions set.
This made me want to introduce a warning for it.

`DeployWarner` is an interface that consumers of this library can choose to implement on top of `DeployObserver`. Therefore this is not breaking.

### Test plan

I did not choose to implement a test because this is only a log output.

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -63,7 +63,7 @@ type DeployObserver interface {
 }
 
 type DeployWarner interface {
-	OnWalkWarning(string)
+	OnWalkWarning(path, msg string)
 }
 
 // DeployOptions holds the option for creating a new deploy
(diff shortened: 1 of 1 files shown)


go/porcelain/deploy.go
```diff
@@ -579,12 +583,16 @@ func bundle(functionDir string, observer DeployObserver) (*deployFiles, error) {
...
 			}
 			functions.Add(file.Name, file)
-		case goFile(filePath, i):
+		case goFile(filePath, i, observer):
 			file, err := newFunctionFile(filePath, i, goRuntime, observer)
 			if err != nil {
 				return nil, err
 			}
 			functions.Add(file.Name, file)
+		default:
```
keiko713, line 592: I'm curious if there is a case that function directory could have some files that are not zip nor js nor go legitimately

mraerino: i think this walks the temporary dir from zisi, therefore only files that zisi picked up should be there

PR #226, opened 2020-04-23
Correct deleteHookBySiteId operationId since it doesn't require site_id

bettse:
As part of work to validate the site_id is passed for hook creation in bitballoon (https://github.com/netlify/bitballoon/issues/5143), I reviewed the open-api and noticed that the delete name was incorrect based on the required parameters.


kitop:
Looks good! Thanks!
Can you make sure to update the go client before we merge? Instructions are in https://github.com/netlify/open-api/blob/master/CONTRIBUTING.md#spec-validation

bettse:
Pushed regenerated golang.  For future reference, I got an error:
>2020/04/23 13:13:10 creating generated file "access_token.go" in "go/models" as definition
failed rendering template data for definition: template execution failed for template definition: template: schematype:2:18: executing "schemaType" at <(len .AllOf) gt 0>: can't give argument to non-function len .AllOf
generate.go:6: running "gobin": exit status 1

which I discovered is because of [stricter type checking in golang 1.14](https://github.com/go-swagger/go-swagger/issues/2215).  I was able to fix it by brew installing `go@1.13` and doing a PATH export.

keiko713:
I'm actually not super familiar with swagger but seems like no-longer-used func/type like `DeleteHookBySiteIDParams` or `DeleteHookBySiteIDNoContent` are not removed. maybe we can remove them?

bettse:
Great catch, the `make generate` can introduce code, but doesn't remove it.  I rm'd the go/plumbling directory and generated again and removed them.

PR #228, opened 2020-04-28
Migrate from Travis to GitHub Action

keiko713:
Several other Netlify open source repositories migrated to GitHub action already, it'd be nice to do so with open-api too.

Note: workflow is pretty much copy&paste from everywhere, any feedback is welcome.


.github/workflows/test.yml
```diff
@@ -0,0 +1,49 @@
+name: Test
+on:
+  # Ensure GitHub actions are not run twice for same commits
+  push:
+    branches: [master]
+    tags: ['*']
+  pull_request:
+    types: [opened, synchronize, reopened]
```
mraerino, line 8: on traffic mesh we are not seeing duplicated runs with just `on: push`: https://github.com/netlify/traffic-mesh/blob/master/.github/workflows/rust.yml#L3

.github/workflows/test.yml
```diff
@@ -0,0 +1,49 @@
...
+
+jobs:
+  test-node:
+    name: Test on node ${{ matrix.node_version }} and ${{ matrix.os }}
```
mraerino, line 12:
```suggestion
    name: Node ${{ matrix.node_version }} on ${{ matrix.os }}
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          node-version: ${{ matrix.node_version }}
+      - name: Git checkout
+        uses: actions/checkout@v1
+      - name: Install dependencies
+        run: npm install
+      - name: Tests
+        run: npm test
+  test-go:
+    name: Test on go ${{ matrix.go_version }} and ${{ matrix.os }}
```
mraerino, line 30:
```suggestion
    name: Go ${{ matrix.go_version }} on ${{ matrix.os }}
```

.github/workflows/test.yml
keiko713: I can confess that I copy&pasted from https://github.com/netlify/js-client/blob/1000eac0fdf9434336adefdde5cce1d46c02d60e/.github/workflows/workflow.yml#L2-L8 , any opinions/suggestions @ehmicky ? (as I copy pasted from you :P)

.github/workflows/test.yml
keiko713: Thanks! yeah realized that the message ending up in checks is pretty duplicated, like this one more 👍

.github/workflows/test.yml
ehmicky: This was suggested by @erezrokah and was originally copied from `netlify-cms`.
When doing only `push`, PRs that come from forked repositories do not seem to be included.
When doing both `push` and `pull_request`, PRs that come from non-forked repositories are run twice, which is the reason behind `branches: [master]`. The `tags` and `types` are just the default values (can be removed).

@erezrokah Please correct me if I'm wrong.

mraerino: ok, that explains why i haven't seen it on a private repo 😄

keiko713: nice 🎉 very good to know, thanks for tips @ehmicky and @erezrokah , sounds like we could copypasta this whenever we migrate other open source repos too

erezrokah: Another thing to consider is that GitHub secrets are not passed to actions triggered from forked repositories.
We had to create a separate testing flow for the CMS project:
https://github.com/netlify/netlify-cms/blob/b5a242ec8ed627e7f5e2ce8af454dd1dda1f71cc/.github/workflows/nodejs.yml#L104

keiko713: good to know, thanks for sharing @erezrokah ! I think it won't be a problem with this repo but I'll keep it in my mind 👍

PR #229, opened 2020-05-07
fix: update dependencies

ldez:
This PR is mainly because of the `github.com/Azure/go-autorest` dependency is not compatible recent version of `github.com/Azure/go-autorest`.
The changes made by `github.com/Azure/go-autorest` about go modules produce is compatibility issue.

I also updated some others dependencies.


keiko713:
Thanks for the contribution, @ldez !
Do you have any reference regarding the incompatibility? I'm not familiar with it so any pointer would be appreciated.

ldez:
The module `github.com/Azure/go-autorest` was divided  in several sub-modules:
- `github.com/Azure/go-autorest/autorest`
- `github.com/Azure/go-autorest/autorest/adal`
- `github.com/Azure/go-autorest/autorest/azure/cli`
- ...

https://github.com/Azure/go-autorest#using-with-go-modules

what's happen when you try to use `open-api` in another project that use a more recent version of `autorest`:

```
go: finding module for package github.com/Azure/go-autorest
github.com/go-acme/lego/v3/providers/dns/test imports
        github.com/Azure/go-autorest: module github.com/Azure/go-autorest@latest found (v14.1.0+incompatible), but does not contain package github.com/Azure/go-autorest
```

keiko713:
ohh makes sense, thank you for the explanation! +1 for updating that part.
I can see that you updated several other dependencies. I'm generally +1 for updating patch version things, but I can see that one package (cenkalti/backoff) is having 2 major version bump, which is concerning to me.
Do you know if that one is not introducing any breaking changes with major version updates? One thing I can see is that v4 requires minimum go 1.12 (which should be okay with this project), but it would be great if you could check that too.

ldez:
The v3 of `backoff` is just the v2 but the right module name: https://github.com/cenkalti/backoff/compare/v2.2.1...v3.0.0

But the v3 contains a bug: https://github.com/go-acme/lego/pull/1043

The v4 works well in Lego and Traefik without any regression.

PR #232, opened 2020-05-09
Add plugin_runs to open-api

vbrown608:
This adds a PluginRun creation endpoint to open-api. This is a Netlify-internal endpoint that's only accessible from within our build system, but needs to be added here in order to included in the Javascript client.

I kept the specification very minimal, especially in terms of acceptable values and required fields, so we can iterate quickly as we approach GA.

I was reading through the redoc docs and saw this line: https://github.com/Redocly/redoc/blob/master/docs/redoc-vendor-extensions.md#usage-in-redoc
> please make sure you add all tags to a group, since a tag that is not in a group, will not be displayed at all!

I think we can use it to our advantage to omit endpoints from the UI. You can verify here: https://deploy-preview-232--open-api.netlify.app

I added an `x-internal` tag, which doesn't do anything but provide a bit of information to readers of the YAML file.

@verythorough I would like to know what you think of this as a short-term solution while we develop our internal API docs approach.


swagger.yml
```diff
@@ -902,6 +902,26 @@ paths:
...
+        - name: plugin_run
+          in: body
+          schema:
+            $ref: '#/definitions/pluginRun'
+      responses:
+        '201':
+          description: CREATED
+          schema:
+            $ref: '#/definitions/pluginRun'
+        default:
```
ehmicky, line 923: Should we add `422`?

swagger.yml
```diff
@@ -2259,6 +2279,25 @@ definitions:
         type: string
       error:
         type: boolean
+  pluginRun:
+    type: object
+    properties:
```
ehmicky, line 2284: We should specify which properties are required.

swagger.yml
```diff
@@ -2259,6 +2279,25 @@ definitions:
...
+      deploy_id:
+        type: string
+      package:
+        type: string
+      version:
+        type: string
+      state:
+        type: string
+      reporting_event:
```
ehmicky, line 2293: Should we add an `enum` there?

swagger.yml
```diff
@@ -2259,6 +2279,25 @@ definitions:
...
+  pluginRun:
+    type: object
+    properties:
+      deploy_id:
+        type: string
+      package:
+        type: string
+      version:
+        type: string
+      state:
```
ehmicky, line 2291: Should we add an `enum` there?

swagger.yml
```diff
@@ -2259,6 +2279,25 @@ definitions:
         type: string
       error:
         type: boolean
+  pluginRun:
```
ehmicky, line 2282: Should we add `description` to each property?
Adding a `x-example` might be good too?

swagger.yml
vbrown608: Skipping because this is an internal only endpoint that may change quickly, per our convo in Slack.

vbrown608:
> @verythorough I would like to know what you think of this as a short-term solution while we develop our internal API docs approach.

Merging so @ehmicky can proceed but let's still talk about it :smiley_cat:

verythorough:
Oh, sorry I missed the question, @vbrown608 ! Nice find in the ReDocs docs. I like your approach as an intermediate step as we work on the long term solution @keiko713 is now organizing.

vbrown608:
Great :+1: 

Now worries on missing, it was more like, I asked the question and then realized I wanted to merge two minutes later :stuck_out_tongue_closed_eyes:

PR #234, opened 2020-05-11
Upgrade go-swagger to 0.23.0

### LINKED ISSUE #233: Code generation fails with Go 1.14
Looks like we're running into this issue: https://github.com/go-swagger/go-swagger/issues/2215

Can be fixed by upgrading to go-swagger v0.23.

vbrown608:
Code generation is failing in Go 1.14, this upgrades go-swagger to fix that.

Changes below are mostly whitespace and rearranging import lines. There's one more I found noteworthy, commented below.

Fixes #233


go/plumbing/operations/operations_client.go
```diff
@@ -9,12 +9,11 @@ import (
...
 
 	"github.com/go-openapi/runtime"
-
-	strfmt "github.com/go-openapi/strfmt"
+	"github.com/go-openapi/strfmt"
 )
 
 // New creates a new operations API client.
-func New(transport runtime.ClientTransport, formats strfmt.Registry) *Client {
+func New(transport runtime.ClientTransport, formats strfmt.Registry) ClientService {
```
vbrown608, line 16: The client is now implemented as an interface so it can be mocked in tests. See https://github.com/go-swagger/go-swagger/issues/1138

PR #235, opened 2020-05-13
Remove deploy_id from PluginRun create body

vbrown608:
Extracts body params of `PluginRun` to a separate scheme that doesn't include `deploy_id`, which should be specified on the path.

This also fixes the version of go-swagger used in the Makefile to match the version used by go.mod.


swagger.yml
```diff
@@ -2299,6 +2297,13 @@ definitions:
         type: string
       text:
         type: string
+  pluginRun:
```
ehmicky, line 2300: The response includes `deploy_id` but not the request body, which explains this, right?

vbrown608: Yep! The response should match the full PluginRun model.

PR #238, opened 2020-05-27
Add 'branch' to createSiteDeploy

RaeesBhatti:
Blocked on: https://github.com/netlify/bitballoon/pull/6224

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -2257,6 +2257,8 @@ definitions:
         type: boolean
       functions:
         type: object
+      branch:
+        type: string
   build:
     type: object
     properties:


swagger.yml
```diff
@@ -2257,6 +2257,8 @@ definitions:
         type: boolean
       functions:
         type: object
+      branch:
```
ehmicky, line 2260: Is this also part of the HTTP response (which is the `deploy` definition I believe)?

RaeesBhatti: No, this only exists for `createSiteDeploy` endpoint

PR #258, opened 2020-09-22
Fix service instance swagger definition

keiko713:
While having a chat in https://github.com/netlify/cli/issues/310#issuecomment-693473685, we realized that the swagger file definition of service instance is incorrect, especially with show/update/delete. These endpoints need to pass `instance_id`.

PR #259, opened 2020-09-24
chore: regenerate go client

erezrokah:
Following up on https://github.com/netlify/open-api/pull/258#issuecomment-698056801.
This regenerates the go client and include changes from:
https://github.com/netlify/open-api/pull/242
https://github.com/netlify/open-api/pull/243
https://github.com/netlify/open-api/pull/247
https://github.com/netlify/open-api/pull/258

I was trying to follow https://github.com/netlify/open-api/blob/master/CONTRIBUTING.md#making-a-new-release, but wasn't sure if to commit these changes directly to the default branch as sometimes they are done via a PR:
https://github.com/netlify/open-api/pull/250/files

Also, following the release process generated two untracked files `swagger_flat.json` and `swagger_go.json`, which I'm not sure if to include or `.gitignore`.


keiko713:
Thanks for following up with this! Totally forgot.

> but wasn't sure if to commit these changes directly to the default branch as sometimes they are done via a PR

I think the best practice here is, if we change swagger.yml, we run the go client regeneration. We all forgot that within all PRs. So having this PR is good 👍 

> Also, following the release process generated two untracked files swagger_flat.json and swagger_go.json, which I'm not sure if to include or .gitignore.

Uhmmm I'm not sure about this honestly. How does that file look like? However, I do think that these don't need to be tracked.

keiko713 (approved):
looking good, thanks!

erezrokah:
> How does that file look like

The files are quite bit (too big to share here). They look similar to the generated `swagger.json` with a slightly different structure.

PR #262, opened 2020-09-29
Add WaitUntilDeployLive Go client method

### LINKED ISSUE #261: Add a WaitUntilDeployLive method to the Go client
Add a method to the Go client porcelain that blocks until the deploy is live, modeled after `WaitUntilDeployReady`.

vbrown608:
Unfortunately, this is really waiting until the deploy is in the `ready` state, but we already have a public method called `WaitUntilDeployReady` that waits until the deploy is `ready` *or* `prepared` (ie ready to start uploading files).

I went with `WaitUntilDeployLive` because it mirrors the "Site is live" message that appears in the deploy logs. Open to other suggestions.

Closes #261 
Part of https://github.com/netlify/pod-the-builder/issues/47
blep

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -311,7 +311,7 @@ func (n *Netlify) waitForState(ctx context.Context, d *models.Deploy, timeout ti
 		context.GetLogger(ctx).WithFields(logrus.Fields{
 			"deploy_id": d.ID,
 			"state":     resp.Payload.State,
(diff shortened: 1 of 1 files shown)


Benaiah (approved):

Benaiah:
Thanks for doing this PR! I agree that "live" is a good term to use in the name for this method, especially given the "Site is live" message. It's also much more specific than "ready", so it'll be clearer what it means when we use it elsewhere.

PR #263, opened 2020-10-06
Add service instances list

keiko713:
A bit of follow-up of https://github.com/netlify/open-api/pull/258
We're missing the endpoint to list service instances for the site in the spec, so adding it.
I also found a bug in my previous PR (status code of `show` was 200 🤦 ) so fixing it here.


vbrown608 (approved):
:carrot:

keiko713:
@erezrokah could you please do the merge + release for this one?

erezrokah:
Released https://github.com/netlify/open-api/releases/tag/v0.18.0. Had to skip `0.17.0` as it was tagged, but not released. Not sure we should do another release just for that tag

PR #264, opened 2020-10-08
Use service as a returned object for showService

keiko713:
It's a small bug, noticed while reviewing https://github.com/netlify/cli/pull/1344

Looks like this one is not currently used for any Netlify owned client (js-client, go client, CLI, etc.).
It's always nice to fix the bug whenever we notice!

### DIFF
--- a/go/plumbing/operations/show_service_responses.go
+++ b/go/plumbing/operations/show_service_responses.go
@@ -51,20 +51,20 @@ func NewShowServiceOK() *ShowServiceOK {
 services
 */
 type ShowServiceOK struct {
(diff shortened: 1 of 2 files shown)


kitop (approved):

erezrokah:
Do you want me to merge and cut a release for this?

keiko713:
@erezrokah thanks for checking! this is small enough and not being a blocker of anything, so this can go with some other release in my opinion (or, of course, we could still release)

erezrokah:
Released as https://github.com/netlify/open-api/releases/tag/v0.18.1

PR #266, opened 2020-10-19
adding refresh_token endpoint

dustincrogers:
#### Description:
Not exactly sure what I am doing here but I believe this will expose the `refresh_token` endpoint thru the open-api spec, making it available to the `js-client` (and I will also regenerate the go client before merging this, as is the protocol I think)

#### Impact:
This could cause downstream issues I think, but is unlikely since it is passing `make validate`, but I am looking for input on that

#### Testing:
Im not sure how to test the client updates for these changes.
passing `make validate`

```
% make validate
GO111MODULE=off go get -u github.com/myitcv/gobin && go mod download
gobin -m -run github.com/go-swagger/go-swagger/cmd/swagger@v0.23.0 validate swagger.yml
2020/10/19 15:50:20
The swagger spec at "swagger.yml" is valid against swagger specification 2.0
```

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -2571,6 +2571,8 @@ definitions:
         type: string
       access_token:
         type: string
+      refresh_token:
+        type: string
       user_id:
         type: string
       user_email:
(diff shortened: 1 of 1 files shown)


dustincrogers:
@keiko713 advised that this is not a proprietary endpoint, but a oauth standard. And we should hardcode the capability into the js-client and go-client instead. Closing

PR #267, opened 2020-11-04
github tools: fix fossa workflow file

tools-netlify:
Updates github actions fossa workflow to use environment files instead of deprecated set-env command.


keiko713 (approved):
🌳

PR #268, opened 2020-11-10
Mark listForms endpoint as deprecated

kitop:
This will no longer be supported and will be retired in 30 days.


vbrown608 (approved):
:wave:

PR #275, opened 2020-12-17
Add getAccountBuildStatus

### LINKED ISSUE #271: Get Build minutes in API
I want to get how many build minutes I have used on my account. but I can’t find in the docs, Can I get build minutes in API ?

vbrown608:
Adds the account build status, including build minutes, to the API docs.

Closes #271


mraerino:
does this match what our private generated open api definition looks like?

rstavchansky (approved):
The very minimalist docs output for this looks fine. In line with what Marcus said, I wondered whether there's a corresponding PR for adding this to the [bitballoon openapi docs](https://bitballoon-openapi.netlify.app/).

swagger.yml
```diff
@@ -2351,6 +2369,39 @@ definitions:
...
+            format: dateTime
+          period_end_date:
+            type: string
+            format: dateTime
+          last_updated_at:
+            type: string
+            format: dateTime
+          included_minutes:
+            type: string
+          included_minutes_with_packs:
```
keiko713, line 2403: I have some minor mixed feeling about adding these "included" two fields here (you can see more in https://github.com/netlify/bitballoon/pull/5881 - tl;dr I didn't wanna add them in this endpoint), but it's less likely to change so... I guess it's okay to be as-is

vbrown608:
Good point, I'll add a corresponding annotation in Bitballoon.

Issue #277, opened 2021-01-04
Golang: Pagination missing


cstaud:
The `ListSites` call does not implement pagination...
I also don't see a chance how to get around this.

By the way: you described the pagination within your documentation at https://docs.netlify.com/api/get-started/#pagination
But the openapi spec seems to miss it.

Am I missing something or does nobody with more than 100 sites use this client?

PR #278, opened 2021-01-07
feat: add pagination support

JGAntunes:
Based on [our docs](https://docs.netlify.com/api/get-started/#pagination) and a couple of expriments I did (and on [bitballoon's codebase](https://github.com/netlify/bitballoon/search?q=pagination)) we support pagination for all of our list endpoints but it's missing from our `open-api` spec.

This PR updates our open-api spec and respective golang client. It also addresses #277 and serves as a base to address https://github.com/netlify/js-client/issues/39, https://github.com/netlify/js-client/pull/44 and https://github.com/netlify/js-client/pull/167.

Let me know what you think 👍


ehmicky (commented):
Finally! :tada: :tada: :tada: 

This looks good to me, but I don't have a good understanding of the API pagination, unlike @vbrown608.

rstavchansky (approved):
Output looks good for docs! 🥇

vbrown608:
Only those controllers that handle `pagination_options` and set `pagination_headers` support pagination. `listAccountTypesForUser` (`AccountsController#types`) does not, for example.

We're in the process of migrating our API docs to be built from inline annotations in the API codebase, so it might be better to add pagination to swagger as needed here.

vbrown608:
I suggest we start with following actions, which all support pagination, and are likely to include 100+ records for some users.
`listSites`
`listSiteSubmissions`
`listSiteDeploys`
`listSiteBuilds`
`listFormSubmissions`
`listAccountAuditEvents`

Thanks for working on this!

JGAntunes:
Thanks for the review 👍 and thanks for checking this @vbrown608! I'll add this to those specific endpoints in that case.

JGAntunes:
Meanwhile I've updated the spec but kept `listSItesForAccount`, let me know if it makes sense to remove that also. 👍

vbrown608 (approved):
Nice catch! Yeah, let's definitely keep that one too. This looks good to me.

PR #281, opened 2021-01-11
Fix createSiteBuildHook and updateSiteBuildHook params

kitop:
We only accept title and branch as params, the rest are only part of the
response.


keiko713 (approved):
looks good! ref: https://github.com/netlify/bitballoon/blob/58d93932f56696c3c913dbe45037a87fece32215/app/controllers/api/v1/build_hooks_controller.rb#L14-L31

Issue #285, opened 2021-01-13
Chore: update release flow to use release PRs


erezrokah:
Our current release flow requires pushing directly to the default branch:
https://github.com/netlify/open-api/blob/master/CONTRIBUTING.md#making-a-new-release

We're in the process of migrating to [release-please](https://github.com/google-github-actions/release-please-action#setting-up-this-action) and should do the same for this repo.
See [here](https://github.com/netlify/framework-info/blob/e17400e915e8d8fb06340867eaf6a9fcd35d9b8c/.github/workflows/release-please.yml#L10) and [here](https://github.com/netlify/framework-info/pull/97).

The blocker for this issue is that we need to find a way to generate the go client (or avoid pushing auto generated files to the repo) during the process of creating the release PR.

erezrokah:
Some more context.
The release please action calls the `release-please` package programmatically:
https://github.com/google-github-actions/release-please-action/blob/88a82e1f9574f0a8bcacf1a13f84ef3d2c6659f3/index.js#L29

If we find the action too limiting we could write a script/our own action/fork the existing action and extend it.
Calling the `release-please` package programmatically will make it easier to manipulate the release PR.

erezrokah:
We could also use the same approach as in:
https://github.com/netlify/cli/blob/491adc49cbc92d93ae0b720a0724fa1824ebb8a4/.github/workflows/verify-docs.yml#L29

To make sure people generate the go client in when submitting a PR.

JGAntunes:
So, I've been looking into how this is tied together and I decided to break this issue into a couple of separate pieces of work, let me know what you think 😄

As it is right now, the versioning of the `open-api` spec is tied together with the `go-client` release - https://github.com/netlify/open-api/blob/master/CONTRIBUTING.md. To some extent, it might make sense, but a problem arises when you want to make changes particular to the client (e.g. https://github.com/netlify/open-api/pull/272). The JS side deals with it by actually encapsulating the `swagger.yml` into it's npm package - https://github.com/netlify/open-api/blob/master/package.json#L47. This issue won't tackle any of these (this is more to provide some context tbh) as I also believe there are some plans to automatically generate the open-api spec (`swagger.yml`) from the API side(?).

So when, releasing, we need to make sure a couple of things happen (not really in a particular order):
- [ ] 1) Bump the version of the `package.json`
- [ ] 2) Bump the version of the `swagger.yml` file to match that of the `package.json`
[...]

PR #288, opened 2021-01-20
chore: add release-please

JGAntunes:
This PR partially addresses #285. The idea is to provide a way to tag, release and bump the versions of both `package.json` and `swagger.yml` files. I tried running some tests on this repo but unfortunately seems like `release-please` will always create PRs against the mainline. As such I've ran some tests yesterday on a small repo - https://github.com/JGAntunes/actions-playground/pull/1 - testing `auto-commit`. The idea here is to rely on it as a workaround to bump the `swagger.yml` file based on `package.json` version (which is initially done by `release-please`).

For this to work I'll need to add a secret to the repo (currently don't have access, maybe you can help me witht it @erezrokah?), similar to what we've done, maybe we can use`netlify-bot`s access token? - https://github.com/netlify/cli/pull/1778#issue-558154582

### DIFF
--- a/.github/workflows/swagger-bump.yml
+++ b/.github/workflows/swagger-bump.yml
@@ -1,5 +1,6 @@
 name: swagger-bump
 on:
+  # Workaround for bumping the swagger.yml file on release-please PRs
   push:
     branches:
       - release-*
@@ -16,4 +17,4 @@ jobs:
       - run: npm version:1-swagger
       - uses: stefanzweifel/git-auto-commit-action@v4
         with:
-          commit_message: "Bump swagger.yml file"
+          commit_message: "chore: bump swagger.yml file"
--- a/.github/workflows/swagger-bump.yml
+++ b/.github/workflows/swagger-bump.yml
@@ -13,7 +13,7 @@ jobs:
         with:
           node-version: 12
           registry-url: 'https://registry.npmjs.org'
-      - run: npm install
+      - run: npm ci
       - run: npm version:1-swagger
       - uses: stefanzweifel/git-auto-commit-action@v4
         with:
--- a/.github/workflows/swagger-bump.yml
+++ b/.github/workflows/swagger-bump.yml
@@ -1,0 +1,19 @@
+name: swagger-bump
+on:
+  push:
+    branches:
+      - release-*
+jobs:
+  swagger-bump:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm install
+      - run: npm version:1-swagger
+      - uses: stefanzweifel/git-auto-commit-action@v4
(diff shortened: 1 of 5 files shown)


erezrokah:
> For this to work I'll need to add a secret to the repo (currently don't have access, maybe you can help me witht it @erezrokah?), similar to what we've done, maybe we can use`netlify-bot`s access token? - [netlify/cli#1778 (comment)](https://github.com/netlify/cli/pull/1778#issue-558154582)

Thanks @JGAntunes I've added the secret. How are we dealing with generating the `go` client?

package.json
```diff
@@ -17,7 +17,6 @@
   "devDependencies": {
     "analytics": "^0.2.0",
     "analytics-plugin-ga": "^0.1.5",
-    "auto-changelog": "^1.16.1",
```
erezrokah, line 20: I think we're missing the `package-lock.json` files changes

.github/workflows/swagger-bump.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  swagger-bump:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm install
```
erezrokah, line 16: Should we use `npm ci` here?

erezrokah (changes requested):
Added a few comments, but will delegate to the Backend the rest of the review.

.github/workflows/swagger-bump.yml
JGAntunes: I thought `npm ci` only installed `dependencies` and not `devDependencies` (which we need here) but now I'm uncertain - https://docs.npmjs.com/cli/v6/commands/npm-ci - and according to the docs doesn't seem like it. I'll change it, thanks @erezrokah 👍

JGAntunes:
> Thanks @JGAntunes I've added the secret. How are we dealing with generating the go client?

I've added a more extensive comment covering that @erezrokah - https://github.com/netlify/open-api/issues/285#issuecomment-763683630

mraerino (commented):
from a process level this looks fine.

this still expects people to manually regenerate the Go client right?

since this has always been the case this is not a blocker: can we think of a process where PRs are denied when the go code is not up-to-date?
that would also block releases that had this problem (but should not if we never merged PRs with failing checks).

erezrokah (approved):

JGAntunes:
> this still expects people to manually regenerate the Go client right?

@mraerino that's correct yes, I've went through a more lengthy boring explanation here - https://github.com/netlify/open-api/issues/285#issuecomment-763683630 - but essentially we can come out with an action to verify if new files are generated through `make all` and haven't been checked in, and fail the build in that case. I would rather have it as a separate PR though. Maybe later on we can improve the automation for generating the go client, but at least that way we would already have some safeguards in place.

mraerino (approved):

PR #290, opened 2021-01-21
fix(package.json): remove scripts no longer required and don't npm ci on swagger bump

JGAntunes:
Currently the [release fails](https://github.com/netlify/open-api/runs/1741730586?check_suite_focus=true) because as part of `npm install/ci` we [run our tests](https://github.com/netlify/open-api/compare/fix/npm-scripts?expand=1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L53) which validate that the versions of the `package.json`  and `swagger.yml` match. I've just realised the `swagger-bump.js` has no external dependencies whatsoever, so running `npm install` isn't required. The alternative would be to drop that specific test from the `prepare` script which I'm not really comfortable with, given it forces us to run the tests before running `npm publish` which we still need to do locally.

It also removes the `prepublishOnly` script as well as the `version:*` scripts which are no longer required.

This should be enough for #289 to go through 🤞

### DIFF
--- a/.github/workflows/swagger-bump.yml
+++ b/.github/workflows/swagger-bump.yml
@@ -13,8 +13,7 @@ jobs:
         with:
           node-version: 12
           registry-url: 'https://registry.npmjs.org'
-      - run: npm ci
-      - run: npm version:1-swagger
+      - run: npm run bump-swagger
       - uses: stefanzweifel/git-auto-commit-action@v4
         with:
           commit_message: "chore: bump swagger.yml file"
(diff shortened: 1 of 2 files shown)


mraerino (approved):
so you don't need to run `git add` anymore?

JGAntunes:
@mraerino nop, [git-auto-commit-action](https://github.com/stefanzweifel/git-auto-commit-action) should take care of that - https://github.com/netlify/open-api/pull/290/files#diff-db93e31c8f3922ed61e6fe01153ed3c3ddd37e75539c63d66659cc9b1bcd0d13L18

erezrokah (approved):

PR #291, opened 2021-01-21
chore(actions): add labeler

JGAntunes:
Use @erezrokah awesome lebeler to tag PRs with the required `type: <type>` label based on the title.


erezrokah (approved):
image

PR #292, opened 2021-01-21
chore(actions): check uncomitted go source code changes

JGAntunes:
This picks up on what has been discussed on #285 and adds a step to verify that the go client is up to date and the files have been checked in (once again based on work @erezrokah did in the CLI 🙏 ). I've tested it in this PR by making some changes to  `swagger.yml` file and not generating a new client - https://github.com/netlify/open-api/runs/1742899676?check_suite_focus=true. - seems solid 😄

I've also added two files to the `.gitignore`, which I believe are only used as [helpers when generating the go glient](https://github.com/netlify/open-api/blob/master/generate.go#L4) and don't need to be checked in. Do tell me if that is not the case though.

### DIFF
--- a/.github/workflows/verify-go-src.yml
+++ b/.github/workflows/verify-go-src.yml
@@ -1,0 +1,33 @@
+name: Verify Go client
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  verify-go:
+    name: Verify go client source code is up to date
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        go_version: [1.14.x]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Install Go ${{ matrix.go }}
+        uses: actions/setup-go@v2.1.3
+        with:
+          go-version: ${{ matrix.go_version }}
+      - name: Checkout code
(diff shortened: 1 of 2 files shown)


keiko713 (approved):
I remember that @erezrokah and I were talking about these swagger_*.json files the other day for what to do. I think it's safe to add them to `.gitignore`, so looks good! 👍

.github/workflows/verify-go-src.yml
```diff
@@ -0,0 +1,33 @@
...
+  pull_request:
+    branches: [master]
+
+jobs:
+  verify-go:
+    name: Verify go client source code is up to date
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        go_version: [1.14.x]
```
mraerino, line 15: latest is `1.15`, do we want to use that?

mraerino (approved):
awesome!

.github/workflows/verify-go-src.yml
JGAntunes: Happy to do it, but maybe in that case we should also add it to our `Test` action - https://github.com/netlify/open-api/blob/master/.github/workflows/test.yml#L34?

PR #297, opened 2021-01-22
Ignore cache directories in root of build directory when deploying

vbrown608:
Fixes https://github.com/netlify/buildbot/issues/1131

Previously, we were moving directories like `node_modules` to the cache after the build was finished but before deploying, so they weren't present at deploy time.

With build plugin steps that run after the site is live, we need to keep these directories around longer. This ignores them from the build directory root during deploy to maintain existing behavior.


vbrown608:
Testing in a production build and looking good.

Benaiah (approved):
LGTM! :tada:

PR #298, opened 2021-02-03
feat!: support flexible cancelation while awaiting deploy state

vbrown608:
Part of https://github.com/netlify/buildbot/issues/1129

Rather than using the build's preprocess timelimit while waiting for a site to go live, we'd like to apply no timelimit other than the overall build timelimit.

Functions that wait for a deploy to enter a certain state take a context, so they don't need to also take a `timelimit` - the time limit can instead be added to the context, and the context's deadline/cancellation can be respected.

The gives users of this package more flexibility, because they can set a time limit on the context, or reuse a context that already has a timeout or cancellation function.

I also pushed setting the default preprocess time limit on `WaitUntilDeployReady` into the caller, `DoDeploy`. I think this is a better fit for the responsibilities of that function and is more symmetrical with `WaitUntilDeployLive`.

### DIFF
--- a/go/porcelain/context/context.go
+++ b/go/porcelain/context/context.go
@@ -3,8 +3,8 @@ package context
 import (
 	"context"
 
-	"github.com/sirupsen/logrus"
 	"github.com/go-openapi/runtime"
+	"github.com/sirupsen/logrus"
 )
 
 type Context interface {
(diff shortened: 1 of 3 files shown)


ingride (approved):
lgtm

PR #304, opened 2021-02-05
docs: example go client

vbrown608:
Adds a package-level doc comment and package-level example to `go/plumbing` demonstrating use of the API client to make a request.

The porcelain package also provides `NewHTTPClient` function that creates a porcelain-layer wrapper client. However, it only supports accessing https://api.netlify.com/api/v1, so it can be used in tests to make requests against a test server.

We could also provide a helper in that package that accepts an endpoint configuration and returns the `porcelain.Netlify` client that wraps that plumbing client. My concern is that package isn't being maintained with a focus on external users lately. It provides wrapper methods for accessing some API endpoint, but not all of them. So, I think it might be better to encourage direct use of the plumbing client for now.

Because it refers to the plumbing client directly, and the porcelain client has a similar API, I put it in the plumbing directory. I have some misgivings about mixing hand-written and generated code there, but it feels important for clarify. I did check that `make generate` doesn't impact those files.


go/plumbing/example_test.go
```diff
@@ -0,0 +1,45 @@
...
+	// For Netlify's production API use DefaultHost, DefaultBasePath, DefaultSchemes
+	transport := httptransport.New(host, DefaultBasePath, []string{"http"})
+	client := New(transport, strfmt.Default)
+
+	// Prepare the API token
+	authInfo := runtime.ClientAuthInfoWriterFunc(func(r runtime.ClientRequest, _ strfmt.Registry) error {
+		r.SetHeaderParam("User-Agent", "Your app")
+		r.SetHeaderParam("Authorization", "Bearer your_netlify_api_token")
+		return nil
+	})
```
vbrown608, line 34: The bearer token can be set more cleanly like below, but I think it's a good idea to encourage API users to set the `User-Agent`.
```
authInfo := httptransport.BearerToken("your_netlify_api_token"))`
```

go/plumbing/doc.go
```diff
@@ -0,0 +1,6 @@
+/*
+Package plumbing provides an auto-generated Go client for accessing Netlify's API.
+
+See https://goswagger.io/generate/client.html for more example usage of an API client generated by go-swagger.
+*/
+package plumbing
```
keiko713, line 6: memo: this doc and the below example doc/code will be shown in https://pkg.go.dev/github.com/netlify/open-api@v1.4.0/go/plumbing (with new version)

keiko713 (approved):
good docs improvement 🙌

vbrown608:
@keiko713 also reminded me to mention, it's going to look like this in the godoc page https://pkg.go.dev/github.com/netlify/open-api@v1.4.0-plumbing-example.4/go/plumbing

I'll do the merge tomorrow.

PR #305, opened 2021-02-05
feat: allow deploy cancellation during file upload

vbrown608:
Sometimes buildbot needs to exit before the deploy is complete, for instance if the build times out of the user cancels it. This PR listens for termination of the passed context and stops trying to upload files if it's done.


go/porcelain/deploy.go
```diff
@@ -383,16 +385,19 @@ func (n *Netlify) uploadFiles(ctx context.Context, d *models.Deploy, files *depl
...
-				wg.Add(1)
-
-				go n.uploadFile(ctx, d, file, observer, t, timeout, wg, sem, sharedErr)
+				select {
+				case sem <- 1:
+					sem <- 1
+					wg.Add(1)
+
+					go n.uploadFile(ctx, d, file, observer, t, timeout, wg, sem, sharedErr)
+				case <-ctx.Done():
```
keiko713, line 394: memo: context cancellation can happen from the buildbot side (e.g. with timeout)

keiko713 (approved):
+1 for the change, but would be really nice if we could check it works with the cancellation in staging or something

PR #306, opened 2021-02-05
chore: improve upload failure logging

vbrown608:
Sometimes we see `Upload cancelled: filename.html` in the logs and it's not clear what happened. This adds more description and includes the `sharedError`, which can hopefully tell us about some bad API response.


keiko713 (approved):
double checked with Vivian that this error message is not shown in build log (aka we can be less worried about wording). looks good 👍

PR #308, opened 2021-02-10
fix: use module v2 in package example

vbrown608:
This code was in master but not the PR branch, so it didn't get the `v2` update all the other files got.


keiko713 (approved):
🥈

PR #311, opened 2021-02-11
fix: don't wrap sharedErr on failed upload

vbrown608:
I thought it would be nice to have the context of the `err` here. However, this `sharedErr` is getting updated by every file attempting to do the upload.

Here's what's happening:
* File 1 fails to upload after some retries with exponential backoff, and sets `sharedErr` saying it failed
* File 2 sees that `sharedErr` isn't nil, so it gives up as well, and wraps the `sharedErr` with its own message
* File 3 sees the *wrapped* `sharedErr` and wraps it again with its own message
* Repeat for files 4 - 100 to create a very ugly error

The causes a deeply nested error and it really doesn't look good. We're logging the original failure like below, so we can see it that way in the logs instead. https://github.com/netlify/open-api/blob/71cc689d4a5e3797563917d9c0e15a9270ecfec0/go/porcelain/deploy.go#L472

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -21,15 +21,14 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
+	apierrors "github.com/go-openapi/errors"
 	"github.com/pkg/errors"
+	"github.com/rsc/goversion/version"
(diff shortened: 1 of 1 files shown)


keiko713 (approved):
:ohno: ➿

PR #313, opened 2021-02-17
fix: don't double-count file uploads in worker pool

vbrown608:
This was receiving from the semaphore twice (`sem <- 1`). Added a test and it's looking much better.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,6 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rsc/goversion v1.2.0
 	github.com/sirupsen/logrus v1.6.0
-	github.com/stretchr/testify v1.6.1
+	github.com/stretchr/testify v1.7.0
 	golang.org/x/net v0.0.0-20201224014010-6772e930b67b // indirect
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -16,6 +16,6 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rsc/goversion v1.2.0
 	github.com/sirupsen/logrus v1.6.0
-	github.com/stretchr/testify v1.6.1
+	github.com/stretchr/testify v1.7.0
```
vbrown608, line 19: Upgraded to get `require.ErrorIs`

mraerino (dismissed):
looks good. the test seems to be very tied to implementation details. would there a way to use more higher-level primitives in the test and reduce the setup boilerplate?

vbrown608:
The higher level primitives we have for creating the client currently only support `https://api.netlify.com/api/v1` as the API endpoint. I did some consolidation of the authorization setup to provide only what's needed for these tests.

keiko713 (approved):
🔄

PR #327, opened 2021-04-18
fix: Switch site update to a put request

wdrach:
Hello! Hope you don't mind a proposed contribution from a lowly open-source developer.

It appears that the request type is incorrect for the updateSite endpoint, although I'm obviously new to this API and can't quite confirm. This is causing problems for me because I'm using the Node API generated by this specification, and the patch request doesn't seem to be having the side-effects that I'm looking for. For example:

Manually updating the custom_domain via the UI makes a PUT request, which works as expected:
Screen Shot 2021-04-18 at 11 15 02 AM

Calling the API function with the same parameters against (theoretically) the same endpoint does not, and I think the incorrect request type might be the culprit:
Screen Shot 2021-04-18 at 11 45 23 AM
Screen Shot 2021-04-18 at 11 23 00 AM

I'm hoping this change in tandem with an API regeneration would fix that problem, but feel free to correct me!

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -111,7 +111,7 @@ paths:
             $ref: '#/definitions/site'
         default:
           $ref: '#/responses/error'
-    patch:
+    put:
       operationId: updateSite
       tags: [site]
       consumes:


erezrokah:
Hi @wdrach can you try:
```js
client.updateSite({ site_id, body: { custom_domain: '...' } })
```

any parameter under the [request body schema](https://open-api.netlify.com/#operation/updateSite) should be included in the `body` param.

This is similar to https://github.com/netlify/js-client#using-openapi-operations

> `PATCH` is generally used to modify a resource and `PUT` is used to replace a resource

wdrach:
That makes sense to me, I'm a bit interested why it's using a different method on the UI than is documented in the open api, but that's above my paygrade.

I'll give this a shot next time I'm fiddling with my project, and reopen if I still have issues. Thanks for the quick response!

PR #328, opened 2021-04-20
fix: remove additional properties on specification

angelinahung:


bryanmikaelian (approved):
Thanks for the contribution!

PR #335, opened 2021-05-03
feat: document latest plugin runs endpoint

vbrown608:
Adds internal-only endpoint to get latest plugin run per package.

Closes https://github.com/netlify/bitballoon/issues/9353


swagger.yml
```diff
@@ -1014,6 +1014,34 @@ paths:
...
+    get:
+      operationId: getLatestPluginRuns
+      tags: [x-internal]
+      description: 'This is an internal-only endpoint.'
+      parameters:
+        - name: site_id
+          in: path
+          type: string
+          required: true
+        - name: packages
```
ehmicky, line 1027: Does the call error when one of the `packages` has no `plugin_run` to return?

swagger.yml
```diff
@@ -1014,6 +1014,34 @@ paths:
...
+          in: query
+          type: array
+          items:
+            type: string
+          required: true
+        - name: state
```
ehmicky, line 1033: Is the default behavior "do not filter by `plugin_run.state`"?

swagger.yml
```diff
@@ -1014,6 +1014,34 @@ paths:
...
+          in: query
+          type: string
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/pluginRun'
```
ehmicky, line 1042: Does the API modify the `plugin_run` when storing it?
Or does it mostly match what `@netlify/build` sent when performing `createPluginRun`?

ehmicky (approved):
:tada:

swagger.yml
vbrown608: yep!

swagger.yml
vbrown608: No, it will just be omitted from the response. So, if no provided package has a `plugin_run` the response will be an empty array.

swagger.yml
vbrown608: We're not currently doing any modification on the backend, it will be returned with the same values that were sent by `@netlify/build`.

swagger.yml
ehmicky: Thanks!

swagger.yml
ehmicky: Great, this should make it easier to assume its shape :+1:

PR #342, opened 2021-06-03
feat: allow setting a build image for a single build

Benaiah:
Allows setting a build image for a single build, using an optional `image` parameter on the `createSiteBuild` endpoint.


rstavchansky (approved):
Approved from Docs

vbrown608 (approved):

PR #353, opened 2021-08-18
feat: added odb support to spec and porcelain

cbosss:
Now uploading ODB functions as part of deploy.


cbosss:
After discussing with team. We are going to defer build/deploy time separation of ODBs until later. See https://netlify.slack.com/archives/C01LYRFHM4L/p1630348781016100 for more context.

PR #366, opened 2021-12-06
feat: add `function_schedules` property

eduardoboucas:
This PR adds support for scheduled functions and can be divided in three parts:

1. https://github.com/netlify/open-api/pull/366/commits/92f58e0bc68f55bd40ef296583463e90c300efba and https://github.com/netlify/open-api/pull/366/commits/13b8f00e3485c364cd4c40a1a101376ebcd8e9c4: Add `function_schedules` as a parameter to the `createSiteDeploy` and `updateSiteDeploy` endpoints, as well as a return property for the deploy object.

2. https://github.com/netlify/open-api/pull/366/commits/3a2316edc053fda29077d258e1df9dcd84e29e90: Add the auto-generated models for the Go client

3. https://github.com/netlify/open-api/pull/366/commits/e86fe3f1e30556ad42bc97868824ced30ecb6f8f: Read function schedules from the functions manifest and passing them as part of the deploy options

Even though the [signature of `bundle()` has changed](https://github.com/netlify/open-api/pull/366/commits/e86fe3f1e30556ad42bc97868824ced30ecb6f8f#diff-602e1ebf353d4906b9de529630317e7f88f5832aa50bbb1ca269741412e1c112R583), this is an internal function and therefore this PR does not constitute a major change.


go/porcelain/deploy.go
```diff
@@ -633,43 +635,51 @@ func bundle(functionDir string, observer DeployObserver) (*deployFiles, error) {
...
 		}
 
 		file, err := newFunctionFile(function.Path, fileInfo, function.Runtime, observer)
 
 		if err != nil {
-			return nil, err
+			return nil, nil, err
+		}
+
+		if function.Schedule != "" {
```
netlify-team-account-1, line 672: today I learned that Golang's JSON unmarshaling uses the empty string as placeholder if a field doesn't exist

netlify-team-account-1 (approved):
:shipit:

go/porcelain/deploy.go
eduardoboucas: That's actually not specific to JSON unmarshalling. When you initialise a variable with a struct type, the properties will use the corresponding [zero values](https://go.dev/tour/basics/12) for their type, which for strings is `""`.

When doing the unmarshalling, any properties that don't exist in the input data will be ignored, so the result will be whatever value was already there. If we wanted to initialise `function.Schedule` with a different value, that value would be the one left in the struct after doing the unmarshalling.

bettse (approved):

klavavej (approved):

PR #368, opened 2022-01-04
feat: add support for scheduled functions to Go client

eduardoboucas:
- Adds a `FunctionSchedules` property to the deploy options when function schedules are detected
- Adds a `scheduled_functions` log field to an existing log message, showing the number of scheduled functions detected for the deploy
- Adds a new log message when a functions manifest file is detected

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -240,8 +240,9 @@ func (n *Netlify) DoDeploy(ctx context.Context, options *DeployOptions, deploy *
 
 	l := context.GetLogger(ctx)
 	l.WithFields(logrus.Fields{
-		"site_id":      options.SiteID,
-		"deploy_files": len(options.files.Sums),
+		"site_id":             options.SiteID,
+		"deploy_files":        len(options.files.Sums),
+		"scheduled_functions": len(schedules),
 	}).Debug("Starting to deploy files")
 	authInfo := context.GetAuthInfo(ctx)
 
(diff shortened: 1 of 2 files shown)


go/porcelain/deploy.go
```diff
@@ -91,7 +91,7 @@ type DeployOptions struct {
 
 	files             *deployFiles
 	functions         *deployFiles
-	functionSchedules []models.FunctionSchedule
+	functionSchedules []*models.FunctionSchedule
```
eduardoboucas, line 94: The change to pointers here is just so that it matches the [`FunctionSchedules` type from `DeployFiles`](https://github.com/netlify/open-api/blob/master/go/models/deploy_files.go#L37) without doing any conversion.

go/porcelain/deploy.go
```diff
@@ -238,10 +238,15 @@ func (n *Netlify) DoDeploy(ctx context.Context, options *DeployOptions, deploy *
 		}
 	}
 
+	if len(schedules) > 0 {
+		deployFiles.FunctionSchedules = schedules
```
eduardoboucas, line 242: This is where we pass the schedules to the deploy options.

go/porcelain/deploy.go
```diff
@@ -580,7 +585,7 @@ func walk(dir string, observer DeployObserver, useLargeMedia, ignoreInstallDirs
 	return files, err
 }
 
-func bundle(functionDir string, observer DeployObserver) (*deployFiles, []models.FunctionSchedule, error) {
+func bundle(ctx context.Context, functionDir string, observer DeployObserver) (*deployFiles, []*models.FunctionSchedule, error) {
```
eduardoboucas, line 588: `bundle` and `bundleFromManifest` are now receiving the context so that they can acquire a logger and print a debug message.

mrdg (approved):

PR #371, opened 2022-01-31
Add missing SRV dnsRecord properties

robotblake:
SRV records contain a `weight` property and a `port` property. These properties are present in the `dnsRecordCreate` definition but not in the `dnsRecord` definition even though they are returned by the API.


kitop:
@robotblake mind rebasing this PR to fix conflicts?

PR #372, opened 2022-02-09
feat: deploy Edge Handlers to internal path

eduardoboucas:


go/porcelain/deploy.go
```diff
@@ -513,6 +525,28 @@ func (n *Netlify) uploadFile(ctx context.Context, d *models.Deploy, f *FileBundl
 	}
 }
 
+func createFileBundle(rel, path string) (*FileBundle, error) {
```
eduardoboucas, line 528: Abstracted into separate function so that it can be used by both `walk` and `walkEdgeHandlers`.

JGAntunes (approved):

eduardoboucas:
Not needed anymore! 🧹

PR #381, opened 2022-06-13
feat: add swagger endpoints and definitions for environment variables

jasonbarry:
This PR adds paths and schemas for endpoints specific to site/account environment variables. 

Screenshot 2022-06-13 at 18 36 03

It adds: 
- 2 new GET requests
- 1 new POST request
- 1 new PUT request
- 2 new DELETE requests

To generate this yaml, I...
- Ran `bin/rake openapi:generate` in BB
- Converted json to yaml
- Copied over relevant paths and schemas
- Converted OAS3 syntax → OAS2 syntax
  - Ensured paths don't start with `/api/v1`
  - Changed $refs to be `#/definitions/thing` instead of `#/components/schemas/thing`
  - Followed https://swagger.io/docs/specification/2-0 and trusted the linter in my REST client 😅 

In terms of testing, I imported this swagger.yml into my REST client, and the new endpoints appeared as expected. I also ran `npm run build` to generate the HTML file, and things looked good there. We should check the Deploy Preview to make sure that the documentation gets generated correctly too: https://deploy-preview-381--open-api.netlify.app/#operation/getEnvVars.


swagger.yml
```diff
@@ -177,6 +177,257 @@ paths:
...
+                      - builds
+                      - functions
+                      - runtime
+                      - post_processing
+                values:
+                  type: array
+                  items:
+                    $ref: '#/definitions/envVarValue'
+        - name: account_id
+          description: Scope response to account_id
```
codebyuma, line 265: [Dust] I didn't think about this in the initial review. Should this be `Scope request` instead of `Scope response`? (for some or all of the commands, I'm not sure 🤔 )

codebyuma (approved):
This looks good to me. I noticed that the general description isn't here that you had [in the bit-balloon version](https://jb-openapispecforenvelopeendpoints--bitballoon-openapi.netlify.app/#tag/Environment-variables), is that intentional? 

`[Beta] Endpoints to retrieve, create, update, and delete environment variables. During the beta period, you must opt in to the new environment variable experience using the Netlify UI before you can use these endpoints to set contextual values and specific scopes.`

swagger.yml
```diff
@@ -177,6 +177,257 @@ paths:
             $ref: '#/definitions/sniCertificate'
         default:
           $ref: '#/responses/error'
+  /accounts/{account_id}/env:
```
lukasholzer, line 180: I'm not an expert on how we treat environment variables but does everyone in a team have access all environment variables from a team?

lukasholzer (approved):

swagger.yml
jasonbarry: Collaborators don't have access to team env vars in the UI. I could be mistaken, but I believe Collaborators have read-only access via CLI/API (so that `netlify dev` can still build the site locally). Collaborators have read/write access to all site-level env vars.

swagger.yml
jasonbarry: Hmm good Q, I think either would be correct. 6 to one, half-dozen to another :-)

aitkenster (dismissed):
🤖

codebyuma:
Just following up on the overview section - since this is the public API docs, are we able to add this note? I noticed none of the other endpoint sections have an overview (even though there's space for it), so I'm not sure if that's a limitation. 

`[Beta] Endpoints to retrieve, create, update, and delete environment variables. During the beta period, you must opt in to the new environment variable experience using the Netlify UI before you can use these endpoints to set contextual values and specific scopes.`

(Sorry if I missed another conversation about this 👀 )

jasonbarry:
Oops, sorry I missed that @codebyuma! I don't think it's possible to add an overview to a group of endpoints -- taking a look at the [swagger docs](https://swagger.io/docs/specification/basic-structure/), I can see how to add a description to the API as a whole (top-level), or to each individual endpoint, but not to a collection of endpoints.

codebyuma:
Thanks for checking! 

At least all the individual env method descriptions start with [beta], so that part is covered 👍  If you haven't opted in via the UI, what kind of response will you get if you try to use these endpoints?

jasonbarry:
@codebyuma good Q! The endpoints would respond in the same way, however the changes wouldn't take effect until the site's `use_envelope` boolean is true. That's a temporary measure until post-beta / GA, when `use_envelope` would be true by default.

codebyuma:
@jasonbarry That's fun 😄 So it would look like you're successfully adding variables with scopes and contextual values for your site to use, but your site actually can't use them? 

It would be nice if the API response could include a message about opting in if that's the case. Not sure how much of an edge case this will be though - I assume most people will opt in using the UI before they start looking at the API... 🤔 

Tagging @melanierichards in case she has any concerns about the missing overview text

melanierichards:
You could consider adding a riff on @codebyuma's caveat to the top of each endpoint. Basically where the [BETA] message is in the screenshot.

codebyuma:
Not sure how it'll look but here's some modified text if you want to try it: 

Longer:
> During the beta period, you must opt in to the new environment variable experience using the Netlify UI before you can use this endpoint. 

Shorter:
> To use this endpoint, opt into the beta environment variable experience using the Netlify UI.

swagger.yml
```diff
@@ -177,6 +177,266 @@ paths:
...
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/envVar'
+        default:
+          $ref: '#/responses/error'
+      description: >-
+        [Beta] Returns all environment variables for an account or site. An
+        account corresponds to a team in the Netlify UI. To use this endpoint, 
+        opt into the beta environment variable experience using the Netlify UI.
```
codebyuma, line 229: [sand] 

Oh no, I'm sorry, I did it again...  I wrote `opt into` instead of `opt in to` in the comment I left for you. I'm going to leave suggested edits for you, hopefully that makes it easier for you to apply 🙏 

```suggestion
        opt in to the beta environment variable experience using the Netlify UI.
```

swagger.yml
```diff
@@ -177,6 +177,266 @@ paths:
...
+    post:
+      tags:
+        - environmentVariables
+      operationId: createEnvVars
+      description: >-
+        [Beta] Creates new environment variables. Granular scopes are available
+        on Pro plans and above. To use this endpoint, opt into the beta environment
```
codebyuma, line 236:
```suggestion
        on Pro plans and above. To use this endpoint, opt in to the beta environment 
```

swagger.yml
```diff
@@ -177,6 +177,266 @@ paths:
...
+      responses:
+        '200':
+          description: OK
+          schema:
+            $ref: '#/definitions/envVar'
+        default:
+          $ref: '#/responses/error'
+      description: >-
+        [Beta] Returns an individual environment variable. To use this endpoint, 
+        opt into the beta environment variable experience using the Netlify UI.
```
codebyuma, line 319:
```suggestion
        opt in to the beta environment variable experience using the Netlify UI.
```

swagger.yml
```diff
@@ -177,6 +177,266 @@ paths:
...
+    put:
+      tags:
+        - environmentVariables
+      operationId: updateEnvVar
+      description: >-
+        [Beta] Updates an existing environment variable and all of its values.
+        Existing values will be replaced by values provided. To use this endpoint, 
+        opt into the beta environment variable experience using the Netlify UI.
```
codebyuma, line 327:
```suggestion
        opt in to the beta environment variable experience using the Netlify UI.
```

swagger.yml
```diff
@@ -177,6 +177,266 @@ paths:
...
+          type: string
+          required: false
+          in: query
+      responses:
+        '204':
+          description: No Content (success)
+        default:
+          $ref: '#/responses/error'
+      description: >-
+        [Beta] Deletes an environment variable. To use this endpoint, opt into the
```
codebyuma, line 402:
```suggestion
        [Beta] Deletes an environment variable. To use this endpoint, opt in to the 
```

swagger.yml
```diff
@@ -177,6 +177,266 @@ paths:
...
+          required: false
+          in: query
+      responses:
+        '204':
+          description: No Content (success)
+        default:
+          $ref: '#/responses/error'
+      description: >-
+        [Beta] Deletes a specific environment variable value. To use this endpoint, 
+        opt into the beta environment variable experience using the Netlify UI.
```
codebyuma, line 439:
```suggestion
        opt in to the beta environment variable experience using the Netlify UI.
```

codebyuma (commented):
Ah, I'm sorry, I made an error in the copy I gave you - I left GH's "suggested edits" to hopefully make it easier for you to apply the changes. 

Once updated, I think this will be good to go! 🙏

codebyuma (approved):
As [discussed on Slack](https://netlify.slack.com/archives/C02TE7L1QUA/p1655482703400669), since changes require the Go client to be regenerated, I'm approving this as-is and we'll update the copy in a follow-up issue. (Just saying this in case anyone else is following along)

PR #383, opened 2022-06-21
chore(docs): add a note about enabling the beta experience in Netlify UI

jasonbarry:
This PR adds an additional note for all the Envelope endpoints to enable the feature in Netlify UI before hitting the endpoints.


codebyuma (approved):
Thanks for making this change! Hopefully it'll help avoid confusion for any users who stumble upon this API documentation before opting in. 

:shipit:

PR #385, opened 2022-07-12
feat: add retry count field

eduardoboucas:
When clients retry an operation, we would like to propagate the retry number to the API, so that it can adjust its behaviour accordingly.

To do this, this PR adds a `X-Nf-Retry-Count` header, which is empty on the first attempt, and contains `1` on the first retry, and so on.

Currently, we're adding the field to the `uploadDeployFunction` only.

Part of https://github.com/netlify/pod-compute/issues/140.

### DIFF
--- a/go/plumbing/operations/upload_deploy_function_parameters.go
+++ b/go/plumbing/operations/upload_deploy_function_parameters.go
@@ -62,6 +62,8 @@ for the upload deploy function operation typically these are written to a http.R
 */
 type UploadDeployFunctionParams struct {
 
+	/*XNfRetryCount*/
+	XNfRetryCount *int64
 	/*DeployID*/
 	DeployID string
 	/*FileBody*/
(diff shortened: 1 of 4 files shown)


mraerino (approved):
does the JS client also need this?

eduardoboucas:
> does the JS client also need this?

Yes 💯! I'll work on that next.

go/plumbing/operations/upload_deploy_function_parameters.go
```diff
@@ -62,6 +62,8 @@ for the upload deploy function operation typically these are written to a http.R
 */
 type UploadDeployFunctionParams struct {
 
+	/*XNfRetryCount*/
+	XNfRetryCount *int64
```
biruwon, line 66: nit: do you really need that big number or something like `uint8` could wokr?

biruwon (approved):

go/plumbing/operations/upload_deploy_function_parameters.go
eduardoboucas: This code is automatically generated by Swagger. It seems to translate `type: integer` into `int64`, and I'm not sure we can configure that.

mraerino: On 64bit system the integer size doesn't make any difference

tarat69 (commented):
You’ve hacked enough

PR #404, opened 2022-10-28
chore: remove beta tags from CLI and docs

pvorozhe:
In the CLI and API docs, remove instances of `[Beta]` from netlify env commands and API reference.

Fixes https://github.com/netlify/pillar-workflow/issues/919 (1/2)

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -315,7 +315,7 @@ paths:
         default:
           $ref: '#/responses/error'
       description: >-
-        Returns an individual environment variable. To use this endpoint, your site
+        Returns an individual environment variable. To use this endpoint, your site 
         must no longer be using the classic environment experience. Migrate now with the Netlify UI.
     put:
       tags:
@@ -323,7 +323,7 @@ paths:
       operationId: updateEnvVar
       description: >-
         Updates an existing environment variable and all of its values.
-        Existing values will be replaced by values provided. To use this endpoint, your site must
+        Existing values will be replaced by values provided. To use this endpoint, your site must 
         no longer be using the classic environment experience. Migrate now with the Netlify UI.
       parameters:
         - name: account_id
@@ -378,7 +378,7 @@ paths:
         - environmentVariables
       operationId: setEnvVarValue
       description: >-
-        Updates or creates a new value for an existing environment variable. To use this endpoint, your site
+        Updates or creates a new value for an existing environment variable. To use this endpoint, your site 
         must no longer be using the classic environment experience. Migrate now with the Netlify UI.
       parameters:
         - name: account_id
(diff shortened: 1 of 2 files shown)


akahn (approved):

swagger.yml
```diff
@@ -224,15 +224,15 @@ paths:
         default:
           $ref: '#/responses/error'
       description: >-
-        [Beta] Returns all environment variables for an account or site. An
+        Returns all environment variables for an account or site. An
         account corresponds to a team in the Netlify UI. To use this endpoint,
         opt in to the beta environment variable experience using the Netlify UI.
```
codebyuma, line 229: I think we should also update the next line that says `...opt in to the beta environment variable experience...` 

I have conflicting thoughts:

a) We could just remove the second line entirely as this new experience is the new default. Similar to how the docs are written as "new experience-first" (similar to how we say "mobile-first"). This is helpful for new teams as they won't need to stop and wonder if they have to migrate, they can just use the API right away.

b) But, we may still find a lot of current teams checking these reference docs and they may have existing sites that need to be migrated first. So, we could replace it with something like: `If you encounter issues with this endpoint, your site might be using the classic environment variables experience. Migrate your site using the Netlify UI`.  

The second option seems safest and it's what we use in the Docs - it's just long... 

I should know this but can't remember - does the API return anything helpful if you try to use these endpoints on a site that hasn't been migrated?

codebyuma: I'm trying to think of a shorter variation that only existing users need to care about: 

`To use this endpoint, your site must no longer be using the classic environment experience. Migrate now with the Netlify UI` ?? 🤔

pvorozhe: Second one sounds reasonable to me - i think there is only so much we can do with so little text to describe this feature rollout

swagger.yml
```diff
@@ -224,17 +224,17 @@ paths:
         default:
           $ref: '#/responses/error'
       description: >-
-        [Beta] Returns all environment variables for an account or site. An
-        account corresponds to a team in the Netlify UI. To use this endpoint,
-        opt in to the beta environment variable experience using the Netlify UI.
+        Returns all environment variables for an account or site. An
+        account corresponds to a team in the Netlify UI. To use this endpoint, your 
+        site must no longer be using the classic environment experience. Migrate now with the Netlify UI.
```
codebyuma, line 229: Sorry, I bet this typo was my fault, I forgot to include the word `variables` here (and this edit should apply to all of the descriptions below):

```suggestion
        site must no longer be using the classic environment variables experience. Migrate now with the Netlify UI.
```

Otherwise this looks good for now 👍  

Curious question, can we include links in these docs? If so, it might be helpful to link to the classic page, so they know what we're talking about. Once it's published, the link will be https://docs.netlify.com/environment-variables/classic-experience/.  If it's not possible, then that's ok.

Something like this?

```suggestion
        site must no longer be using the <a href="https://docs.netlify.com/environment-variables/classic-experience/">classic environment variables experience</a>. Migrate now with the Netlify UI.
```

codebyuma (approved):
The copy update looks good to me - glad that the link works! Thanks for making these updates

PR #415, opened 2023-01-18
chore(docs): point to new environment variable endpoints in site crud operation descriptions

jasonbarry:
This PR updates the description of site CRUD operations to include a message that new environment variable endpoints should be used instead of relying on `build_settings.env` and `repo.env`. 

These are the operations affected by this change: 

- [createSite](https://open-api.netlify.com/#tag/site/operation/createSite)
- [createSiteInTeam](https://open-api.netlify.com/#tag/site/operation/createSiteInTeam)
- [getSite](https://open-api.netlify.com/#tag/site/operation/getSite)
- [listSites](https://open-api.netlify.com/#tag/site/operation/listSites)
- [listSitesForAccount](https://open-api.netlify.com/#tag/site/operation/listSitesForAccount)
- [updateSite](https://open-api.netlify.com/#tag/site/operation/updateSite)

Closes https://github.com/netlify/pillar-workflow/issues/628


jasonbarry:
@codebyuma what do you think of the language I introduced here? Happy to update it to something else!

codebyuma (dismissed):
These descriptions look good to me 👍  

Is there any way to add something to catch people's eyes when they look at the docs? I assume we can't use an icon but what about bold or italicized text at the start, something like `deprecation warning:`? Or is that too much? 🤔 

Additional reviewer note: I was going to ask if we need to make it clear that you need to migrate first but we note that for each of the new endpoints anyway and you see that when you click through to those docs. So I think it's ok not to mention that here.

jasonbarry:
What do you think @melanierichards, should we add `Deprecation warning:` to the start of each?

swagger.yml
```diff
@@ -74,6 +74,9 @@ paths:
               $ref: '#/definitions/site'
         default:
           $ref: '#/responses/error'
+      description: >-
+        Environment variable keys and values inside `build_settings.env` and `repo.env` will soon be removed.
```
melanierichards, line 78: I wonder if this message might make some folks panic that their env vars will be completely removed from the platform? I wonder if we might prevent such a misread, e.g.

> Environment variable keys and values will soon be moved from `build_settings.env` and `repo.env` to a new service.

melanierichards:
> What do you think @melanierichards, should we add `Deprecation warning:` to the start of each?

I prefer the suggestion to render these in bold text, but assume we don't have the ability to do so? I do think it's possible this could get missed, but `Deprecation warning:` might be a little strong given we're not deprecating the entire endpoint.

Maybe there's a softer prefix? Like `Key deprecation warning:` or `Important update:`? Probably a good option that I haven't thought of yet...

jasonbarry:
How's **Note:** for a prefix? I've added it in bold.

jasonbarry:
@melanierichards how's this?

melanierichards (approved):
LGTM, thank you!

PR #418, opened 2023-01-20
feat: add functions_config property to deployfiles

khendrikse:
Adding a new functions_config property to deployfiles for 

https://github.com/netlify/pod-compute/issues/156

and connected to https://github.com/netlify/bitballoon/pull/13556

### DIFF
--- a/go/models/deploy_files.go
+++ b/go/models/deploy_files.go
@@ -38,6 +38,9 @@ type DeployFiles struct {
 
 	// functions
 	Functions interface{} `json:"functions,omitempty"`
+
+	// functions config
+	FunctionsConfig interface{} `json:"functions_config,omitempty"`
 }
 
 // Validate validates this deploy files
--- a/go/plumbing/operations/operations_client.go
+++ b/go/plumbing/operations/operations_client.go
@@ -675,7 +675,7 @@ func (a *Client) CreateServiceInstance(params *CreateServiceInstanceParams, auth
 }
 
 /*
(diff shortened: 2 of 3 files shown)


go/models/deploy_files.go
```diff
@@ -38,6 +38,9 @@ type DeployFiles struct {
 
 	// functions
 	Functions interface{} `json:"functions,omitempty"`
+
+	// functions config
+	FunctionsConfig interface{} `json:"functions_config,omitempty"`
```
Skn0tt, line 43: it feels a bit whacky to put `interface{}` here, because that's a very nondescript type - but it seems like all other fields do the same :shrug: if others feel similar, maybe we can open an issue to make this a little more precise.

Skn0tt (approved):
lgtm

eduardoboucas:
I was too late, but I think we should type the functions config object at the API spec level as opposed to leaving it up to the implementation level.

Swagger makes it a bit awkward to type objects with dynamic keys, but I _think_ we could do something like:

```yaml
functionConfig:
  type: object
  properties:
    display_name:
      type: string
    something_else:
      type: boolean

functionsConfig:
  type: object
  additionalProperties: 
    type:
      $ref: '#/definitions/functionConfig'
```

khendrikse:
@eduardoboucas I'll open another PR ☺️ I tried this and couldn't get it to work before, but I think I may have now!

khendrikse:
@eduardoboucas is this what you meant? https://github.com/netlify/open-api/pull/419

eduardoboucas:
> @eduardoboucas is this what you meant? #419

Yes, exactly! ✨

PR #420, opened 2023-01-30
feat: add functions_config parameter to deploy

khendrikse:
Some work to support https://github.com/netlify/pod-compute/issues/192 in buildbot :)

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -95,6 +95,7 @@ type DeployOptions struct {
 	files             *deployFiles
 	functions         *deployFiles
 	functionSchedules []*models.FunctionSchedule
+	functionsConfig   map[string]models.FunctionConfig
 }
 
 type uploadError struct {
@@ -225,7 +226,7 @@ func (n *Netlify) DoDeploy(ctx context.Context, options *DeployOptions, deploy *
 
 	options.files = files
 
-	functions, schedules, err := bundle(ctx, options.FunctionsDir, options.Observer)
+	functions, schedules, functionsConfig, err := bundle(ctx, options.FunctionsDir, options.Observer)
 	if err != nil {
 		if options.Observer != nil {
 			options.Observer.OnFailedWalk()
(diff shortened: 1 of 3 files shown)


go/porcelain/deploy.go
```diff
@@ -95,6 +95,7 @@ type DeployOptions struct {
 	files             *deployFiles
 	functions         *deployFiles
 	functionSchedules []*models.FunctionSchedule
+	functionsConfig   map[string]models.FunctionConfig
```
biruwon, line 98: based on the previous properties, files, functions and functionSchedules are not required options, but if I am not wrong, now functionsConfig will be required to be at least an empty map?

khendrikse: It was not meant as being required. But are you saying this because of the usage of `*` that the previous ones were considered not required? :)

khendrikse: 🤔 yeah it's really new for me as well. I thought you could make it work like this as well because the map should not throw an error if it's not available. And I think in the swagger file, it was not set up as being required, so I'm thinking maybe it will use that as well to know wether or not the field itself should be required or not. But maybe @eduardoboucas can shine some light on this as well :)

go/porcelain/deploy.go
```diff
@@ -255,11 +257,16 @@ func (n *Netlify) DoDeploy(ctx context.Context, options *DeployOptions, deploy *
...
+	if options.functionsConfig != nil {
+		deployFiles.FunctionsConfig = options.functionsConfig
+	}
+
 	l := context.GetLogger(ctx)
 	l.WithFields(logrus.Fields{
 		"site_id":             options.SiteID,
 		"deploy_files":        len(options.files.Sums),
 		"scheduled_functions": len(schedules),
+		"functions_config":    functionsConfig,
```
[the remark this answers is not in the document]
khendrikse: Makes total sense now that you say it, I removed it, as I think right now it is not useful to log something. Maybe when we get the autogenerated information in the future, that would be interesting to log. To know how many of the functions were autogenerated functions for example :)

khendrikse:
Confirmed that this works by trying it out in a draft pr of buildbot! :sparkles:

Skn0tt (approved):
LGTM!

PR #425, opened 2023-02-16
fix: upload a single file per sha

4xposed:
Addresses the `open-api` part of this issue: https://github.com/netlify/netlify-server/issues/2626

### DIFF
--- a/go/porcelain/deploy.go
+++ b/go/porcelain/deploy.go
@@ -92,6 +92,8 @@ type DeployOptions struct {
 
 	Observer DeployObserver
 
+	SkipSameFileOptimization bool
+
 	files             *deployFiles
 	functions         *deployFiles
 	functionSchedules []*models.FunctionSchedule
@@ -327,8 +329,16 @@ func (n *Netlify) DoDeploy(ctx context.Context, options *DeployOptions, deploy *
 		return deploy, nil
 	}
 
-	if err := n.uploadFiles(ctx, deploy, options.files, options.Observer, fileUpload, options.UploadTimeout); err != nil {
-		return nil, err
+	// Temporary way to ensure safe rollout of skip uploading files with same digest optimization
+	// We'll
+	if options.SkipSameFileOptimization {
+		if err := n.uploadFilesWithSkipSameFileOptimization(ctx, deploy, options.files, options.Observer, fileUpload, options.UploadTimeout); err != nil {
+			return nil, err
+		}
+	} else {
+		if err := n.uploadFiles(ctx, deploy, options.files, options.Observer, fileUpload, options.UploadTimeout); err != nil {
+			return nil, err
+		}
 	}
 
 	if options.functions != nil {
@@ -384,7 +394,7 @@ func (n *Netlify) WaitUntilDeployLive(ctx context.Context, d *models.Deploy) (*m
 	return n.waitForState(ctx, d, "ready")
 }
 
-func (n *Netlify) uploadFiles(ctx context.Context, d *models.Deploy, files *deployFiles, observer DeployObserver, t uploadType, timeout time.Duration) error {
+func (n *Netlify) uploadFilesWithSkipSameFileOptimization(ctx context.Context, d *models.Deploy, files *deployFiles, observer DeployObserver, t uploadType, timeout time.Duration) error {
 	sharedErr := &uploadError{err: nil, mutex: &sync.Mutex{}}
 	sem := make(chan int, n.uploadLimit)
 	wg := &sync.WaitGroup{}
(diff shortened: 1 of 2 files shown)


JGAntunes:
Added a couple of things we've chatted about @4xposed.

Added a flag to `DoDeploy` in order to shift between our optimization and the current upload setup. That way we should be able to safely roll this out from Buildbot's side. I was also thinking that this should also suffice in case anyone is relying on this lib's API, without breaking any compatibility.

Also added test coverage for function uploads.

Let me know what you think 👍

4xposed (commented):
LGTM!!

though GH does not let me approve since I'm an author in this PR

JGAntunes (dismissed):
LGTM. Paired on this work with @4xposed, we're both happy with it 👍

go/porcelain/deploy.go
```diff
@@ -327,8 +329,16 @@ func (n *Netlify) DoDeploy(ctx context.Context, options *DeployOptions, deploy *
 		return deploy, nil
 	}
 
-	if err := n.uploadFiles(ctx, deploy, options.files, options.Observer, fileUpload, options.UploadTimeout); err != nil {
-		return nil, err
+	// Temporary way to ensure safe rollout of skip uploading files with same digest optimization
+	// We'll
```
jobala, line 333: You had more to say 😅 or this is the whole comment?

go/porcelain/deploy.go
```diff
@@ -384,6 +394,57 @@ func (n *Netlify) WaitUntilDeployLive(ctx context.Context, d *models.Deploy) (*m
 	return n.waitForState(ctx, d, "ready")
 }
 
+func (n *Netlify) uploadFilesWithSkipSameFileOptimization(ctx context.Context, d *models.Deploy, files *deployFiles, observer DeployObserver, t uploadType, timeout time.Duration) error {
```
jobala, line 397: What do you think of using `type` instead of `t` for the variable name?

go/porcelain/deploy.go
```diff
@@ -384,6 +394,57 @@ func (n *Netlify) WaitUntilDeployLive(ctx context.Context, d *models.Deploy) (*m
...
+	for _, sha := range required {
+		if files, exist := files.Hashed[sha]; exist {
+			for range files {
+				count++
+			}
+		}
+	}
+
+	log := context.GetLogger(ctx)
+	log.Infof("Uploading %v files", count)
```
jobala, line 421: Will this count be greater than the files we actually upload? Since we upload the first file instead of all files that belong to a `sha`.

https://github.com/netlify/open-api/blob/fd8e4cb7f089086e99b9b4906d67fa5eb9f4c248/go/porcelain/deploy.go#L423-L425

go/porcelain/deploy.go
```diff
@@ -384,6 +394,57 @@ func (n *Netlify) WaitUntilDeployLive(ctx context.Context, d *models.Deploy) (*m
...
+			select {
+			case sem <- 1:
+				wg.Add(1)
+				go n.uploadFile(ctx, d, file, observer, t, timeout, wg, sem, sharedErr)
+			case <-ctx.Done():
+				log.Info("Context terminated, aborting file upload")
+				return errors.Wrap(ctx.Err(), "aborted file upload early")
+			}
+
+			if len(files) > 1 {
```
jobala, line 436: Is this logged to the customer? If it is, I wonder how useful it is to the customer to know we skipped some files during upload.

jobala:
Thinking Out Loud

Why do we have multiple files with the same SHA? Should we do this optimisation where we are adding files to the same SHA?

go/porcelain/deploy.go
JGAntunes: `type` is a reserved word in Go I believe. Unsure how well it would play with its compiler + I would think it could potentially be a bit confusing 🤔

go/porcelain/deploy.go
JGAntunes: Yes it could be, this represents all the file refs we're going to upload. It just happens that some don't actually need it (and we log those accordingly).

go/porcelain/deploy.go
JGAntunes: No, all the logs in the Go client use the sys logger:
- https://github.com/netlify/buildbot/blob/c87074660cd9a4568c0ac759bc083987e94204b3/api/netlify_api_client.go#L111

JGAntunes:
> Why do we have multiple files with the same SHA? Should we do this optimisation where we are adding files to the same SHA?

@jobala - https://github.com/netlify/netlify-server/issues/2626#issuecomment-1424511349 and https://github.com/netlify/buildbot/pull/2559 - have more context. We're not adding files to the same SHA. What happens during the deploy process is that the client calculates file digests (sha1's) for the files it wants to upload. From the server we get a reply with a list of the file digests (sha1's for files and sha2's for functions) that we need to upload. What we're doing here is instead of making multiple requests to upload that same file content multiple times, doing it only once (which is what the server also expects). More info in - https://docs.netlify.com/api/get-started/#deploy-with-the-api

go/porcelain/deploy.go
JGAntunes: Fixed in - https://github.com/netlify/open-api/pull/425/commits/552dc0d5e54296b399ff1dfa6093d17d9ee84ab6

JGAntunes:
Moving this to draft so that it doesn't get merged by accident.

JGAntunes:
This has been fully rolled out via a RC in our Buildbot. https://github.com/netlify/open-api/pull/425/commits/b405b15d419daabc1ab26b51e489f6c6372b5cb9 - takes care of removing the conditional parameter and the duplicated method. Once this is realased (as a patch) we can bump our client in Buildbot.

@4xposed mind re-reviweing the PR and if you're 👍 we can go ahead, merge this and update the client package in buildbot.

4xposed (commented):
LGTM (I cannot approve it though, since I opened it 🙃 )

PR #428, opened 2023-02-27
chore(file-upload): log file size when uploading a file

JGAntunes:
Related with - https://github.com/netlify/pillar-workflow/issues/1126 - adding a field to our debug log line with the size of the file we're uploading.


kitop (approved):
⚖️

PR #432, opened 2023-03-03
fix: add legal compliance footer

MarcL:
This adds the legal compliance footer as discussed in https://github.com/netlify/pod-ecosystem-frameworks/issues/418.

It's not pretty but it injects the footer into the generated `redoc` page in the same way as the meta tags in the `<head>`.

## Screenshot

CleanShot 2023-03-06 at 14 33 56


jasonbarry:
Thanks for jumping on this @MarcL to make sure that we are compliant!! 

I wanted to point out this resource for modifying footer content for redoc-powered sites: https://redocly.com/docs/developer-portal/guides/navbars-footers/#modify-footer-contents

Maybe this approach can be used instead of the injection approach? 

Not a blocker by any means, esp with the Mar 7 turnaround time, but wanted to throw that out there!

MarcL:
Hey @jasonbarry.

I did look at this as I really don't like the way this is done. Injected the code in this way didn't sit right with me!

It felt beyond the scope of the work here as the link you shared would need us to also create the developer portal I think. Currently, it uses the Swagger schema to generate and it looked like we'd need more config. I had a read through the `redoc-cli` docs and I couldn't immediately see how we use the `siteConfig.yaml` approach that they suggest.

Are we ok with this as is for now and we can look to improve the docs in a future PR?
Or if you can see how to use the developer portal approach then happy to update to use it if you can point me in the right direction.

Thanks!

jasonbarry:
@MarcL yeah, it does seem out of scope for this PR. I think we'd need to start by not [shelling out to redoc-cli](https://github.com/netlify/open-api/blob/master/src/docs/build.js#L18-L19), and instead just call it from an npm script to read a yml config file. But we can save that for another day!

akahn (dismissed):

akahn (approved):

MarcL:
Pushed some minimal tweaks to CSS to match the open API docs font.

PR #444, opened 2023-04-18
chore: test with go 1.20 and cleanup tooling

### LINKED ISSUE #416: Remove gobin in makefile in favor of plain go
There's a step in the makefile that depends on gobin, which has been archived as of Nov 8 of last year. We should use the normal go binary instead of depend on gobin.

danez:
Added go 1.20 to the test matrix.

This gets rid of `gobin`, as it is not needed anymore, instead `go run` can be used. Also `richgo` is removed (for fancy test output) because I haven't seen this in any of our other repos, and even the author recommends not using it :)

I had to update `swagger-go` from 0.23 to 0.24, because 0.23 wouldn't work with `go run`. 0.24 introduces [case-insensitive enums](https://github.com/go-swagger/go-swagger/pull/2264/files), hence the changes to the generated models. I tried updating to the latest version (0.30), but then all generated models are very different, if we want to update we should maybe do it in steps.

Also remove the deprecated labeler job.

Fixes #416


.github/workflows/test.yml
```diff
@@ -33,10 +33,10 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        go_version: [1.17.x, 1.18.x, 1.19.x]
+        go_version: [1.17.x, 1.18.x, 1.19.x, 1.20.x]
```
mraerino, line 36: i think you can remove `1.17` and maybe also `1.18` as i think it would be fine to support 2 recent versions only

.github/workflows/test.yml
```diff
@@ -33,10 +33,10 @@ jobs:
...
     runs-on: ${{ matrix.os }}
     steps:
-      - name: Install Go ${{ matrix.go }}
+      - name: Install Go ${{ matrix.go_version }}
```
mraerino, line 39: nice catch!

mraerino (approved):
nice! nothing blocking

.github/workflows/test.yml
danez: Thanks to https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions which also checks for errors.

.github/workflows/test.yml
danez: Okay changed.

go.mod
```diff
@@ -10,13 +10,9 @@ require (
 	github.com/go-openapi/strfmt v0.19.11
 	github.com/go-openapi/swag v0.19.12
 	github.com/go-openapi/validate v0.20.0
-	github.com/go-swagger/go-swagger v0.23.0
```
mraerino, line 13: i think you need to keep this dependency so that the `go run` for the `go:generate` instructions work.

the usual advice on how to do this is by creating a package called `tools` that has this:

```go
package tools

import (
  _ "github.com/go-swagger/go-swagger"
)
```

danez: if `go run pkg@version` is used then the do module is compiled and run without it being installed or affecting the current module. That is what I did. If you prefer I can change that back to what you are suggesting.

mraerino: > if `go run pkg@version` is used then the go module is compiled and run without it being installed or affecting the current module. That is what I did. If you prefer I can change that back to what you are suggesting.

ah good point. i'd prefer having it in `go.mod` so renovate can update it though

danez: Done

mraerino (approved):

PR #466, opened 2023-06-06
feat: add feature_flags parameter to the getSite endpoint

sarahetter:
Adds to the swagger documentation the ability to pass in a `feature_flags` parameter to the `getSite` api, which will return a list of feature flags and whether or not they are enabled. 

Unsure about the values, but went off of this definition: https://github.com/netlify/bitballoon/pull/15495/files#diff-c7977f571d39c09cf8480243e53e44eb1c40acfc7fcb1d76446222e28de96478R100

Also unsure about how to mark as `x-internal`, as elsewhere in the swagger, only endpoints are set with ` tags: [x-internal]`, not parameters.

Depends on: https://github.com/netlify/bitballoon/pull/15495
Related: https://github.com/netlify/pod-dev-foundations/issues/519


mraerino:
do we really want allow customer to look at the feature flags? iirc we intentionally obfuscate flag names in the frontend to not leak new features etc. if this really needs to be in a openapi doc, consider using [the private docs](https://www.notion.so/netlify/Bitballoon-OpenAPI-specification-1cfb6159c5a94d89b0bfd0840dff1ed9)

sarahetter:
@mraerino I've updated to the newer version of redoc cli (the version we were using was deprecated) - and now this allows for us to filter out  properties & methods with `x-internal:true` as described here: https://redocly.com/docs/cli/guides/hide-apis/, while keeping the swagger.yml with all the correct info for the api. 

I'm also seeing that the "Verify Go client" job is failing - I know nothing about Go - what do I need to do here to be able to run `make all`?

erikw (approved):
`x-internal`s are not visible in the deploy preview 👍 

https://deploy-preview-466--open-api.netlify.app/#tag/site/operation/getSite

Issue #476, opened 2023-07-18
Don't retry file uploads with `422` responses


JGAntunes:
Following up on - https://github.com/netlify/pillar-support/issues/625

We should not be retrying file/function uploads which return 422. See [here](https://github.com/netlify/netlify-server/blob/main/server/request_file_upload.go#L101) for a better idea of the response codes we return.

erikw:
Related: same idea for functions upload
https://github.com/netlify/functions-origin/issues/358

JGAntunes:
Feels like this is already being worked on by dev foundations - https://github.com/netlify/open-api/pull/478 - going to close it 👍

PR #478, opened 2023-07-20
Feat: Don't retry function upload on 400 or 422 status

erikw:
~# WIP
Don't retry function upload on 4xx. A 4xx response~

When a function upload receives a 400 or 422 status code, we know that retrying the upload will fail. This change returns early rather than retrying.


go/porcelain/deploy.go
```diff
@@ -538,10 +538,17 @@ func (n *Netlify) uploadFile(ctx context.Context, d *models.Deploy, f *FileBundl
...
+			if ok {
+				if apiErr.Code() == 401 {
+					sharedErr.mutex.Lock()
+					sharedErr.err = operationError
+					sharedErr.mutex.Unlock()
+				}
+
+				// TODO this changes retry behaviour for the fileUpload case as well. OK?
+				if apiErr.Code()/100 == 4 {
+					return nil
```
JGAntunes, line 550: I know this is a draft PR so maybe the implementation isn't finished yet but we'll need to set the `sharedErr` too here.

go/porcelain/deploy.go
```diff
@@ -538,10 +538,17 @@ func (n *Netlify) uploadFile(ctx context.Context, d *models.Deploy, f *FileBundl
...
-				sharedErr.mutex.Unlock()
+			if ok {
+				if apiErr.Code() == 401 {
+					sharedErr.mutex.Lock()
+					sharedErr.err = operationError
+					sharedErr.mutex.Unlock()
+				}
+
+				// TODO this changes retry behaviour for the fileUpload case as well. OK?
+				if apiErr.Code()/100 == 4 {
```
JGAntunes, line 549: I would voice the same concern as @biruwon pointed out here. I'm concerned that retrying all the 4xx calls might be a stretch? (i.e. should we retry 429's too?)

JGAntunes (commented):
The direction looks right 👍

Something I would suggest we had is a property (a boolean flag of some sort) that we can use to enable this new behaviour and that way feature flag it on Buildbot's side.

go/porcelain/deploy.go
jenae-janzen: Updated to just skip the retry for 400 and 422 status codes 🙂

jenae-janzen:
> Something I would suggest we had is a property (a boolean flag of some sort) that we can use to enable this new behaviour and that way feature flag it on Buildbot's side.

I added this to the `options` struct, but am also open to passing it in as a parameter. I started adding it in [Buildbot here](https://github.com/netlify/buildbot/pull/2881), but it's not set up quite right yet and I haven't set up the flag in DevCycle yet.

jenae-janzen:
This is ready for review. We set `skipRetry` to `false` by default if no value is received for the feature flag. The [accompanying change](https://github.com/netlify/buildbot/pull/2881) is also ready for review, but depends on this change going out first because this is where `options` is defined.

go/porcelain/deploy.go
```diff
@@ -344,12 +345,14 @@ func (n *Netlify) DoDeploy(ctx context.Context, options *DeployOptions, deploy *
 		return deploy, nil
 	}
 
-	if err := n.uploadFiles(ctx, deploy, options.files, options.Observer, fileUpload, options.UploadTimeout); err != nil {
+	skipRetry := options.SkipRetry || false
```
eduardoboucas, line 348: You don't need `|| false` here. By saying that `options.SkipRetry` is a boolean, it'll get initialised with [the zero value for that type](https://go.dev/tour/basics/12), which is `false`. So you can pass it directly to `uploadFiles`.

go/porcelain/deploy.go
```diff
@@ -83,7 +83,8 @@ type DeployOptions struct {
 	BuildDir          string
 	LargeMediaEnabled bool
 
-	IsDraft bool
+	IsDraft   bool
+	SkipRetry bool
```
eduardoboucas, line 87: Is the idea that buildbot will set this property based on the value of a feature flag? If so, I imagine we'll remove this once we finish the rollout of the flag?

This only controls the retry mechanism for functions, not regular files, so it might be worth to name the variable accordingly, or at the very least add a comment explaining what it's for.

jenae-janzen: Is it worth using this feature flag for both function _and_ file uploads, since we don't want to retry 400/422 file uploads either? Here's the issue for that: https://github.com/netlify/open-api/issues/476

As written currently, this does change the behavior for file uploads. Erik mentioned it in [a comment in his original draft](https://github.com/netlify/open-api/pull/478/commits/2f75b17dec49ef7d09c3126c28d69268ddba52f9#diff-602e1ebf353d4906b9de529630317e7f88f5832aa50bbb1ca269741412e1c112R548). 

If we want to keep the scope of this flag/pr to _just_ handle function uploads for now, that's totally fine too, we would just change [this line ](https://github.com/netlify/open-api/pull/478/files#diff-602e1ebf353d4906b9de529630317e7f88f5832aa50bbb1ca269741412e1c112R350) from:
```
n.uploadFiles(ctx, deploy, options.files, options.Observer, fileUpload, options.UploadTimeout, skipRetry)
```
to 
```
n.uploadFiles(ctx, deploy, options.files, options.Observer, fileUpload, options.UploadTimeout, false)
```

eduardoboucas: Ah, my bad! I was reviewing this on my phone last night and the diff made it seem that the check for `skipRetry` was under the `case functionUpload:` branch. Ignore me.

If stopping retries for files is also something we want to do, then that makes sense!

go/porcelain/deploy.go
```diff
@@ -538,10 +541,19 @@ func (n *Netlify) uploadFile(ctx context.Context, d *models.Deploy, f *FileBundl
...
+					sharedErr.mutex.Lock()
+					sharedErr.err = operationError
+					sharedErr.mutex.Unlock()
+				}
+
+				if skipRetry && (apiErr.Code() == 400 || apiErr.Code() == 422) {
+					sharedErr.mutex.Lock()
+					sharedErr.err = operationError
+					sharedErr.mutex.Unlock()
+					return nil
```
JGAntunes, line 555: So I had to spend some time looking through this but I believe there's some room for improvement here (in the retry strategy I mean 😅)

I was initially going to say that we didn't want to `return nil` here but then I got that the idea is to actually exit the retry early. We have a check at the beggining of the retry method where we error out if `sharedErr` is set:
- https://github.com/netlify/open-api/blob/99d733ef7085a95798fb2a151e89d775cbc68bab/go/porcelain/deploy.go#L493-L496

However this case is also unideal because it just means we'll keep on retrying and erroring out here until we exhaust the count.

I was reading through the docs for `backoff` - https://pkg.go.dev/github.com/cenkalti/backoff/v4#Retry - and it seems like they have support for a `PermanentError` - https://pkg.go.dev/github.com/cenkalti/backoff/v4#PermanentError - which is a way for us to exit out of the retry loop while keeping the "error semantics" correct. It seems like this would be the perfect fit for this case? What do you think?

go/porcelain/deploy_test.go
```diff
@@ -287,7 +287,7 @@ func TestUploadFiles_Cancelation(t *testing.T) {
 	for _, bundle := range files.Files {
 		d.Required = append(d.Required, bundle.Sum)
 	}
-	err = client.uploadFiles(ctx, d, files, nil, fileUpload, time.Minute)
+	err = client.uploadFiles(ctx, d, files, nil, fileUpload, time.Minute, false)
```
JGAntunes, line 290: Wondering if we could add a test here which exercises the new code paths we introduced (i.e. checking we don't retry `400` and `422` errors when uploading).

JGAntunes (commented):
Left some comments, let me know what you all think 👍

go/porcelain/deploy_test.go
jrwhitmer: @JGAntunes do you have any pointers for the testing? I seem to be a bit stuck here. We are also having trouble running the test suite locally, which is slowing this down a bit 😬

go/porcelain/deploy.go
```diff
@@ -401,8 +404,9 @@ func (n *Netlify) WaitUntilDeployLive(ctx context.Context, d *models.Deploy) (*m
 	return n.waitForState(ctx, d, "ready")
 }
 
-func (n *Netlify) uploadFiles(ctx context.Context, d *models.Deploy, files *deployFiles, observer DeployObserver, t uploadType, timeout time.Duration) error {
+func (n *Netlify) uploadFiles(ctx context.Context, d *models.Deploy, files *deployFiles, observer DeployObserver, t uploadType, timeout time.Duration, skipRetry bool) error {
 	sharedErr := &uploadError{err: nil, mutex: &sync.Mutex{}}
+	permanentErr := &backoff.PermanentError{Err: nil}
```
4xposed, line 409: It's a bit unclear for me why we instantiate `permanentErr` here, as could you elaborate?

as far as I know `backoff.PermanentError` is to be used as return from a `backoff.Retry()` block

go/porcelain/deploy.go
```diff
@@ -538,10 +542,16 @@ func (n *Netlify) uploadFile(ctx context.Context, d *models.Deploy, f *FileBundl
...
-				sharedErr.mutex.Unlock()
+			if ok {
+				if apiErr.Code() == 401 {
+					sharedErr.mutex.Lock()
+					sharedErr.err = operationError
+					sharedErr.mutex.Unlock()
+				}
+
+				if skipRetry && (apiErr.Code() == 400 || apiErr.Code() == 422) {
+					operationError = permanentErr
```
4xposed, line 553: I might be wrong, but I think we should wrap the existing error in a `backoff.PermanentError{}` (for which `backoff` provides the `Permanent()` function:

```suggestion
					operationError = backoff.Permanent(operationError)
```

go/porcelain/deploy_test.go
4xposed: about running the test suite locally, if it seems like nothing is running it's probably because of the retries loop.

try running the tests as:
```
 go test -race github.com/netlify/open-api/v2/go/porcelain --run TestUploadFiles400Errors -v
```

And you should get some output (hopefully)

jenae-janzen: Thanks for the tip @4xposed!! We got the tests running and can see the output which has been very helpful :D

go/porcelain/deploy.go
jenae-janzen: Initially we weren't sure how to use PermanentError, but your other comment pointed us in the right direction I think

go/porcelain/deploy.go
jenae-janzen: Ah, I think that does make sense. And then when `backoff` receives the `Permanent` error, it should stop retrying, if I understand correctly.

go/porcelain/deploy_test.go
```diff
@@ -317,9 +317,38 @@ func TestUploadFiles_Errors(t *testing.T) {
...
 	require.Equal(t, err.Error(), "[PUT /deploys/{deploy_id}/files/{path}][500] uploadDeployFile default  &{Code:0 Message:}")
 }
+func TestUploadFiles400Errors(t *testing.T) {
+	ctx := gocontext.Background()
+
+	server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {
+		rw.WriteHeader(http.StatusUnprocessableEntity)
```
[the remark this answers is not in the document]
jenae-janzen: This is super helpful and definitely makes sense! -- looking at other test cases I see this is done similarly. However, when we set ```rw.Write([]byte(`{"message": "Unprocessable Entity", "code": 422}`))``` we keep getting the error`&{0 } (*models.Error) is not supported by the TextConsumer, can be resolved by supporting TextUnmarshaler interface`.

So then the apiError isn't set and then our test skips over the code we're trying to test and keeps retrying 🙈

Looking at other examples in other tests, it seems like it works to set the response this way when calling `client.Operations.UploadDeployFile`, but not when calling `client.uploadFiles` -- when calling `uploadFiles` the error is thrown from `UploadDeployFile`.

We've been trying to debug this by writing the body different ways, trying to marshal the response etc, but haven't had any success yet 🤔 Is there maybe something silly we're missing?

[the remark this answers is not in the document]
jenae-janzen: This is where the error is being thrown from: https://github.com/netlify/open-api/blob/master/go/plumbing/operations/operations_client.go#L4131

go/porcelain/deploy_test.go
```diff
@@ -317,10 +317,113 @@ func TestUploadFiles_Errors(t *testing.T) {
...
+	// files, err := walk(dir, nil, false, false)
+	// require.NoError(t, err)
+	// d := &models.Deploy{}
+	// for _, bundle := range files.Files {
+	// 	d.Required = append(d.Required, bundle.Sum)
+	// }
+	// // Set SkipRetry to true
+	// err = client.uploadFiles(ctx, d, files, nil, fileUpload, time.Minute, true)
+
+	// // Function upload:
```
jenae-janzen, line 361: Tried testing both file uploads and function uploads but the error was the same

go/porcelain/deploy_test.go
```diff
@@ -317,10 +317,113 @@ func TestUploadFiles_Errors(t *testing.T) {
...
+	files, err := walk(dir, nil, false, false)
+	require.NoError(t, err)
+	d := &models.Deploy{}
+	for _, bundle := range files.Files {
+		d.Required = append(d.Required, bundle.Sum)
+	}
+	// Set SkipRetry to false
+	err = client.uploadFiles(ctx, d, files, nil, fileUpload, time.Minute, false)
+	// require.Equal(t, err, "[PUT /deploys/{deploy_id}/files/{path}][422] uploadDeployFile default  &{Code:422 Message: Unprocessable Entity}")
+	require.Equal(t, attempts, 12)
```
jenae-janzen, line 424: The number of attempts doesn't seem to be consistent, so maybe there's a better way to assert this

go/porcelain/deploy_test.go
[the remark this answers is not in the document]
4xposed: I think It might be the order on which those are set, mockserver is a bit quirky at times. try if setting the headers in this order helps:

```go
rw.Header().Set("Content-Type", "application/json; charset=utf-8")
rw.WriteHeader(http.StatusUnprocessableEntity)
rw.Write([]byte(`{"message": "Unprocessable Entity", "code": 422 }`))
```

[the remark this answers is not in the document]
jenae-janzen: 👋 @4xposed thanks so much for the tip! That worked and now as far as I can tell, the server is responding as expected. However, now we're hitting a different problem and struggling to debug it. 

We now receive the error with the correct code and message, but when we get to [this line](https://github.com/netlify/open-api/pull/478/files#diff-602e1ebf353d4906b9de529630317e7f88f5832aa50bbb1ca269741412e1c112R542)
```apiErr, ok := operationError.(apierrors.Error)```
no matter what we try, apiErr is always `nil` and `ok` is always false. 

`operationError` is an `error` that looks like:
```error(*github.com/netlify/open-api/v2/go/plumbing/operations.UploadDeployFunctionDefault) *{_statusCode: 422, Payload: *github.com/netlify/open-api/v2/go/models.Error {Code: 422, Message: "Unprocessable Entity"}}```

In my debugger, I can access the status code by calling either `operationError._statusCode` or `operationError.Payload.Code`, but not by calling `operationError.(apierrors.Error)`. However, I can't call `operationError.Payload.Code` in the actual code, because `Payload`/`Code` isn't defined on the Error type. 

So since `operationError.(apierrors.Error)` never seems to evaluate to anything and there wasn't a test for the original code, we're uncertain if we're doing something wrong here, or if the original code was actually working as intended. 

Are we missing something here? Is there a better way to test this (perhaps non-locally?) 

cc @jrwhitmer @JGAntunes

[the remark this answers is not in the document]
4xposed: Hey @jenae-janzen, I have been looking into this and I think this casting to `apierrors.Error` never actually worked 😅 

I opened this PR to fix that issue: https://github.com/netlify/open-api/pull/486

once that's merged you should be able to rebase and resume your work on this PR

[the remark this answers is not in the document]
JGAntunes: @jenae-janzen this has been merged - https://github.com/netlify/open-api/pull/486 - I think if we rebase the errors should now behave as expected 👍

>  or if the original code was actually working as intended.

Spoiler alert, it never worked as intended 😂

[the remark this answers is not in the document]
4xposed: @jenae-janzenI believe you will be in Europe soon-ish, if you want to pair on this feel free to book something in my calendar!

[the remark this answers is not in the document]
jenae-janzen: Pairing might not be necessary since most of this seems to be working, but I'd appreciate you taking another look when you have the time 🙂

go/porcelain/deploy_test.go
jenae-janzen: @4xposed this is one other tiny detail that sometimes makes the test fail - every once in a while it retries 13 times 🤔

go/porcelain/deploy_test.go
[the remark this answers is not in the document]
4xposed: we could call `err.Unwrap()` seems like `backoff.PermanentError` has a few methods for this: https://github.com/cenkalti/backoff/blob/v4/retry.go#L129

I don't think it's strictly necessary, but it would be nice to at least have some way that we have the _right_ error and not _any_ error, if that makes sense.

What do you think?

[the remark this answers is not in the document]
jenae-janzen: I'll give it a shot! I do agree it would be nice to have the right error if possible

go/porcelain/deploy_test.go
4xposed: yes I _think_ (emphasis on _think_) it retries infinitely until the `context` times out or it's cancelled.

I think checking that there's more than 1 attempt (or some arbitrary number that's not too high) might be enough, maybe?

I will try to look more into the retry mechanism tomorrow morning to try to figure out if we can have more control in the test of how many times we retry

jenae-janzen: I added a check to just make sure that the attempts are greater than 1 :) That should probably be sufficient for what we're testing

go/porcelain/deploy_test.go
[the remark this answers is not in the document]
jenae-janzen: Okay, I used `Require. ErrorContains` instead and that seemed to do the trick for verifying we have the correct error message

jrwhitmer (approved):

PR #483, opened 2023-08-22
feat: support function route `methods` field

Skn0tt:
Part of https://github.com/netlify/pod-dev-foundations/issues/579, based on https://github.com/netlify/open-api/pull/481.

### DIFF
--- a/go/models/function_route.go
+++ b/go/models/function_route.go
@@ -6,8 +6,13 @@ package models
 // Editing this file might prove futile when you re-run the swagger generate command
 
 import (
+	"encoding/json"
+	"strconv"
+
+	"github.com/go-openapi/errors"
 	"github.com/go-openapi/strfmt"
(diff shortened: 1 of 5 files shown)


klavavej (commented):
I don't see any new prose so not sure this really needs docs review

list of methods in swagger.yml looks right to me though I'm not sure where to see this in the api reference - I poked around in the deploy preview and couldn't find the new content

Skn0tt:
Agreed, I don't think docs review is necessary for this change per se - but apparently, you're a codeowner for the file.

I had to search really deep in the reference - the change is located in https://deploy-preview-483--open-api.netlify.com/#tag/deploy/operation/createSiteDeploy, buried under a couple more collapsibles:
Screenshot 2023-08-25 at 07 46 25

eduardoboucas (approved):

PR #485, opened 2023-09-04
Feat: Add endpoint to move to ready

biruwon:
Will move deploy to ready state when in validation


swagger.yml
```diff
@@ -1076,6 +1076,22 @@ paths:
...
+    post:
+      operationId: readyDeploy
+      tags: [deploy]
+      parameters:
+        - name: deploy_id
+          type: string
+          in: path
+          required: true
+      responses:
+        '201':
```
biruwon, line 1089: Not sure why on [the BB side](https://github.com/netlify/bitballoon/pull/16569/files#diff-c6ed00a289567dfc399f1032e78413680734d0cb65230b37df98d72e615f4663R368) we are saying a 200 for this and the rest of operations that update the deploy state

swagger.yml
```diff
@@ -1076,6 +1076,22 @@ paths:
             $ref: '#/definitions/deploy'
         default:
           $ref: '#/responses/error'
+  /deploys/{deploy_id}/ready:
```
4xposed, line 1079: I'm not sure if it has already been discussed but to me this does not seem very RESTful.

What would you think to use the `update` endpoint `PUT /deploys/{deploy_id}` with a payload of `{state: "READY"}`

we could defer by the payload what to do, and not only seems a bit more RESTful it's a bit more representative of the change we want to do with the resource (Deploy)

I wonder what the rest of the team would think of this as well.

cc. @JGAntunes @kitop @julienp

kitop: Moving this convo to https://github.com/netlify/bitballoon/pull/16569/files#r1315114299

swagger.yml
kitop: I don't quite follow. 200 seems the correct response in most cases. Could be a copy-pasta error in others? Seems like we should respect 200 here, right?

biruwon: It seems to be the case, updated ready, cancel and restore endpoint to match what we really return, as 201 should be when a resource it's created.

kitop:
Closing this based on the chats we've been having.

PR #488, opened 2023-09-08
feat: send `runtime` and `runtimeVersion` separately to `uploadDeployFunction`

Skn0tt:
Part of https://github.com/netlify/pod-dev-foundations/issues/581.

Currently, we're sending on parameter `runtime` that's either `js` or `go`, depending on the type of function. But it can also be something like `nodejs14.x`, if that was written into the `runtimeVersion` field of `manifest.json`. This mangles together two pieces of information, that we'd better have separately:

- `runtime` - which language is being used?
- `runtimeVersion` - what's the AWS Lambda image that should be used to deploy?

It's useful to have these separately, e.g. right now where we're migrating functions on the `go` runtime from the Lambda runtime `go1.x` (soon to be deprecated) to `provided.al2`.

This PR splits up the `runtime` field into these two fields. It also sets the runtimeVersion for individual go files to `provided.al2`.

Depends up updates to our API to be made first.


Skn0tt:
Closing favour of an approach where we reuse the `runtime` field: https://github.com/netlify/bitballoon/pull/16659

PR #490, opened 2023-09-21
feat: upload go binaries as al2-compatible zip

Skn0tt:
Ensures that Go binaries that have `provided.al2` as their `runtimeVersion` are uploaded in a `provided.al2`-compatible ZIP. This means that the name of the binary inside the created ZIP needs to be `bootstrap`, and that we're sending `runtime=provided.al2` to the Netlify API.

Part of https://github.com/netlify/pod-dev-foundations/issues/581.


mraerino (commented):
how are we making sure that binaries we get are ready for this runtime? they need to use a certain version of the Lambda Go SDK right?

Skn0tt:
The binary needs to be built with `GOOS=linux GOARCH=amd64`: https://docs.aws.amazon.com/lambda/latest/dg/golang-package.html#heading:r1i:
This hasn't changed between the `go1.x` and `provided.al2` runtimes, so our customers won't need to change anything to the binaries they're uploading.

go/porcelain/deploy_test.go
```diff
@@ -507,11 +516,13 @@ func TestBundleWithManifest(t *testing.T) {
 	assert.Equal(t, "hello-js-function-test", schedules[0].Name)
 	assert.Equal(t, "* * * * *", schedules[0].Cron)
 
-	assert.Equal(t, 2, len(functions.Files))
+	assert.Equal(t, 3, len(functions.Files))
 	assert.Equal(t, "a-runtime", functions.Files["hello-js-function-test"].Runtime)
 	assert.Empty(t, functions.Files["hello-js-function-test"].FunctionMetadata.InvocationMode)
 	assert.Equal(t, "some-other-runtime", functions.Files["hello-py-function-test"].Runtime)
 	assert.Equal(t, "stream", functions.Files["hello-py-function-test"].FunctionMetadata.InvocationMode)
+	assert.Equal(t, "provided.al2", functions.Files["hello-go-binary-function"].Runtime)
```
Skn0tt, line 524: If there's a `runtimeVersion: "provided.al2"` in the manifest, we now pick that up and send it to the API as the runtime.

Skn0tt:
I've made some updates to go along with https://github.com/netlify/zip-it-and-ship-it/pull/1609, and would love another review.

JGAntunes (approved):
LGTM. CI seems to be complaining but I don't know how relevant it is since we've changed the `porcelain` part and nothing on the auto generated bit - https://github.com/netlify/open-api/actions/runs/6494408840/job/17671324967?pr=490#step:5:10

PR #496, opened 2023-09-26
feat: Remove asset optimization properties

kitop:
Related to https://www.netlify.com/blog/deprecation-of-post-processing-asset-optimization-feature/

On October 17th, we stopped accepting the properties for `js`, `css`, and `images` under `processing_settings`. This change removes them from the docs entirely to avoid confusion.


JGAntunes (approved):
🧹

PR #497, opened 2023-09-28
Add `buildbot_version` and `build_version` to notifyBuildStart

Skn0tt:
Adds two parameters that we can use to pass version information from buildbot into bitballoon. See https://github.com/netlify/bitballoon/pull/16714#issuecomment-1738921918 for more context.

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -1200,6 +1200,14 @@ paths:
         type: string
         in: path
         required: true
+      - name: buildbot_version
+        type: string
+        in: query
+        required: false
+      - name: build_version
+        type: string
+        in: query
+        required: false
     post:
       operationId: notifyBuildStart
       tags: [build]


swagger.yml
```diff
@@ -1200,6 +1200,14 @@ paths:
         type: string
         in: path
         required: true
+      - name: buildbot_version
+        type: string
+        in: query
```
kitop, line 1205: Shouldn't this be in the body? (Same with the one below)

Skn0tt: It's in the query intentionally, so we can spot it in our logs (you can never have enough info!). What'd be the benefit from putting it into the body?

kitop: I think it's a bit nicer and cleaner being a `POST` request, specially if we think of passing more data later. But not a strong opinion, we can always change it if we need to.

Skn0tt: Since these aren't required, and they're more of a meta information piece than the subject of the `POST`, I prefer keeping them in `query` 🤷 Let's go with this for now :)

kitop (approved):

PR #501, opened 2023-10-11
feat: purge API endpoint docs

paulo:
Basic documentation for the purge API.

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -719,6 +719,25 @@ paths:
             $ref: '#/definitions/file'
         default:
           $ref: '#/responses/error'
+  /purge:
+    post:
+      operationId: purgeSiteCache
+      tags: [purge]
+      consumes:
+        - application/json
+      parameters:
+        - name: payload
+          in: body
+          schema:
+            $ref: '#/definitions/purge'
+          required: true
+      responses:
+        '202':
+          description: OK
+        '400':
+          description: Invalid request parameters
+        '404':
+          description: Site not found
   /sites/{site_id}/snippets:
     parameters:
       - name: site_id
@@ -2901,6 +2920,21 @@ definitions:
         type: string
       goal_position:
         type: string
+  purge:
+    type: object
+    properties:
+      site_id:
+        type: string
+      site_slug:
+        type: string
+      domain:
+        type: string
+      deploy_alias:
+        type: string
+      cache_tags:
+        type: array
+        items:
+          type: string
   deploy:
     type: object
     properties:
@@ -3685,6 +3719,7 @@ x-tagGroups:
       - environmentVariables
       - file
       - metadata
+      - purge
       - snippet
   - name: Domain names
     tags:
@@ -3749,6 +3784,8 @@ tags:
     x-displayName: Metadata
   - name: snippet
     x-displayName: Snippet
+  - name: purge
+    x-displayName: Purge cache
   - name: dnsZone
     x-displayName: DNS zone
   - name: sniCertificate


swagger.yml
```diff
@@ -719,6 +719,25 @@ paths:
             $ref: '#/definitions/file'
         default:
           $ref: '#/responses/error'
+  /purge:
+    post:
+      operationId: purgeSiteCache
+      tags: [purge]
+      consumes:
+        - application/json
```
mraerino, line 727: technically we also support formdata payloads. do we want to doc that?

mrdg (dismissed):

swagger.yml
```diff
@@ -719,6 +719,25 @@ paths:
...
+      parameters:
+        - name: payload
+          in: body
+          schema:
+            $ref: '#/definitions/purge'
```
mraerino, line 732: idea, what if we had this so we can specify the required paramters better?
```suggestion
            oneOf:
              - $ref: '#/components/schemas/PurgeSiteID'
              - $ref: '#/components/schemas/PurgeSiteSlug'
              - $ref: '#/components/schemas/PurgeDomain'
```

paulo: talked about it on slack, we're running open-api 2.0 still, which doesn't support this

klavavej (commented):
It would be nice to include descriptions for the method and parameters similar to how we include descriptions for env var methods and parameters here https://deploy-preview-501--open-api.netlify.com/#tag/environmentVariables

Based on https://www.notion.so/netlify/Documentation-Guidelines-Bitballoon-OpenAPI-Endpoints-781dcabfe7ca46e29b826cd1ddfa6b97 I think you might have to add/update controller and serializer files for this

Here's an annotated screenshot from that notion page to help show what I mean about adding descriptions
Screenshot 2023-10-12 at 8 12 15 AM

paulo:
> Based on https://www.notion.so/netlify/Documentation-Guidelines-Bitballoon-OpenAPI-Endpoints-781dcabfe7ca46e29b826cd1ddfa6b97 I think you might have to add/update controller and serializer files for this

Correct me if I'm wrong, but that's for the internal bitballoon docs, while we're documenting the public API, right?

I've tried adding that documentation to the body params, but looks like only path and query parameters support descriptions in open-api 2.0 :( I've added a description to the operation anyway.

go/plumbing/operations/operations_client.go
```diff
@@ -3402,7 +3402,7 @@ func (a *Client) ProvisionSiteTLSCertificate(params *ProvisionSiteTLSCertificate
 }
 
 /*
-PurgeCache purge cache API
+PurgeCache Purges cached content from Netlify's CDN. Support purging by Cache-Tags.
```
klavavej, line 3405: [pebble] suggestion to tidy up grammar and match response header name verbatim for clarity 

```suggestion
PurgeCache Purges cached content from Netlify's CDN. Supports purging by Cache-Tag.
```

swagger.yml
```diff
@@ -722,6 +722,8 @@ paths:
   /purge:
     post:
       operationId: purgeCache
+      description: >-
+        Purges cached content from Netlify's CDN. Support purging by Cache-Tags.
```
klavavej, line 726: [pebble] suggestion to tidy up grammar and match response header name verbatim for clarity

```suggestion
Purges cached content from Netlify's CDN. Supports purging by Cache-Tag.
```

klavavej (approved):
> Correct me if I'm wrong, but that's for the internal bitballoon docs, while we're documenting the public API, right?

Hmmmm, I was under the impression that the guidelines were for a new customer-facing API reference that we migrated to over the last couple years. But I do see now that when I load https://bitballoon-openapi.netlify.app/ linked in the docs it's marked internal so maybe the effort to make the new reference external fizzled and now it's internal only? I'm not sure

At any rate I think the addition you made is helpful and I have a small suggestion for tidying it up a bit

klavavej (approved):
docs changes LGTM

PR #509, opened 2023-10-27
fix: al2 uploads need to have function name in SHA

Skn0tt:
There's a detailed writeup of why this is necessary in https://linear.app/netlify/issue/COM-95/bug-what-happens-when-we-upload-two-functions-with-the-exact-identical.

In short: Netlify requires two different functions to have separate hashsums. Before, this was the case because every ZIP contained the function name as the entrypoint filename. With the changes in #505, this will now always be `bootstrap`, so it doesn't effect the hashsum anymore. By adding the filename as a comment, we get back this behaviour.


kitop (approved):
Nice one!

go/porcelain/deploy.go
```diff
@@ -979,6 +979,7 @@ func createHeader(archive *zip.Writer, i os.FileInfo, runtime string) (io.Writer
 		return archive.CreateHeader(&zip.FileHeader{
 			CreatorVersion: 3 << 8,     // indicates Unix
 			ExternalAttrs:  0777 << 16, // -rwxrwxrwx file permissions
+			Comment:        i.Name(),
```
JGAntunes, line 982: Tiny one, but could we add a comment here so that future us are not 🤯 trying to figure out why?

Skn0tt: yes! done in [22d8ab9](https://github.com/netlify/open-api/pull/509/commits/22d8ab916d0baf122c9a71f35a8349bba759f269)

mraerino (approved):

PR #515, opened 2023-11-03
feat: add `prefer_static` to functions config

eduardoboucas:
Part of https://linear.app/netlify/issue/COM-32/functions-v2-inline-config-does-not-respect-cdn-shadowing.


JGAntunes (approved):

klavavej (approved):
I'm not sure where to see the docs update in https://deploy-preview-515--open-api.netlify.com/ but the updated swagger.yaml looks reasonable to me.

We'll need a corresponding update in the[ Get started with functions doc under "Route requests"](https://docs.netlify.com/functions/get-started/?fn-language=ts#route-requests) I can open a PR for that on Friday unless you beat me to it

PR #516, opened 2023-11-06
feat: add getSiteEnvVars endpoint

charliegroll:
Adding this alias endpoint so you can get site env vars without knowing the account slug/id at request time.


swagger.yml
```diff
@@ -307,6 +307,48 @@ paths:
...
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/envVar'
+        default:
+          $ref: '#/responses/error'
+      description: Returns all environment variables for a site.
```
jasonbarry, line 351: I'd add a sentence like 

```
This convenience method behaves the same as `getEnvVars`, but doesn't require an `account_id` as input.
```

mraerino (approved):

jasonbarry (commented):
Thanks so much @charliegroll!! Can we also use this PR as an opportunity to update the descriptions  of other env var methods? I.e. a find-and-replace of 

from 
```
Environment variable keys and values will soon be moved
```

to 
```
Environment variable keys and values have moved
```

swagger.yml
charliegroll: addressing in https://github.com/netlify/open-api/pull/518

PR #522, opened 2024-01-03
refactor: replace promisify with fs/promises version on convert.js

hereje:
refactor: replace make-dir with fs/mkdir
    
- replace functionality related to create a directory
- remove dependency make-dir no longer needed

Related to https://github.com/netlify/cli/issues/3941

### DIFF
--- a/src/convert.js
+++ b/src/convert.js
@@ -11,7 +11,7 @@ const JSON_OUTPUT = `${OUTPUT_DIR}/swagger.json`
 const convertOpenApi = async function () {
   const [openapiDef] = await Promise.all([
     SwaggerParser.validate(YAML_INPUT, { dereference: { circular: false } }),
-    mkdir(OUTPUT_DIR, { recursive: true }, () => {}),
+    mkdir(OUTPUT_DIR, { recursive: true }),
   ])
   const openapiJson = JSON.stringify(openapiDef, null, 2)
   await writeFile(JSON_OUTPUT, openapiJson)
--- a/src/convert.js
+++ b/src/convert.js
@@ -1,9 +1,6 @@
-const { writeFile, mkdir } = require('fs')
-const { promisify } = require('util')
+const { writeFile, mkdir } = require('fs/promises')
 
 const SwaggerParser = require('swagger-parser')
-
-const pWriteFile = promisify(writeFile)
 
 const YAML_INPUT = `${__dirname}/../swagger.yml`
 const OUTPUT_DIR = `${__dirname}/../dist`
@@ -17,7 +14,7 @@ const convertOpenApi = async function () {
     mkdir(OUTPUT_DIR, { recursive: true }, () => {}),
   ])
   const openapiJson = JSON.stringify(openapiDef, null, 2)
-  await pWriteFile(JSON_OUTPUT, openapiJson)
+  await writeFile(JSON_OUTPUT, openapiJson)
 }
 
 convertOpenApi()
--- a/src/convert.js
+++ b/src/convert.js
@@ -1,8 +1,7 @@
-const { writeFile } = require('fs')
+const { writeFile, mkdir } = require('fs')
 const { promisify } = require('util')
 
 const SwaggerParser = require('swagger-parser')
-const makeDir = require('make-dir')
 
 const pWriteFile = promisify(writeFile)
 
@@ -15,7 +14,9 @@ const JSON_OUTPUT = `${OUTPUT_DIR}/swagger.json`
 const convertOpenApi = async function () {
   const [openapiDef] = await Promise.all([
     SwaggerParser.validate(YAML_INPUT, { dereference: { circular: false } }),
-    makeDir(OUTPUT_DIR),
+    mkdir(OUTPUT_DIR, { recursive: true }, () => {
+      console.log(`${OUTPUT_DIR} created`)
+    }),
   ])
   const openapiJson = JSON.stringify(openapiDef, null, 2)
   await pWriteFile(JSON_OUTPUT, openapiJson)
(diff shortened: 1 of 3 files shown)


src/convert.js
```diff
@@ -15,7 +14,9 @@ const JSON_OUTPUT = `${OUTPUT_DIR}/swagger.json`
 const convertOpenApi = async function () {
   const [openapiDef] = await Promise.all([
     SwaggerParser.validate(YAML_INPUT, { dereference: { circular: false } }),
-    makeDir(OUTPUT_DIR),
+    mkdir(OUTPUT_DIR, { recursive: true }, () => {
+      console.log(`${OUTPUT_DIR} created`)
```
mraerino, line 18: why did you add this log line?

mraerino: i don't see any reason for logging here, so i'd rather remove it

src/convert.js
```diff
@@ -15,7 +14,7 @@ const JSON_OUTPUT = `${OUTPUT_DIR}/swagger.json`
 const convertOpenApi = async function () {
   const [openapiDef] = await Promise.all([
     SwaggerParser.validate(YAML_INPUT, { dereference: { circular: false } }),
-    makeDir(OUTPUT_DIR),
+    mkdir(OUTPUT_DIR, { recursive: true }, () => {}),
```
kitop, line 17: Shouldn't this be using the promise version of `mkdir`? https://nodejs.org/api/fs.html#fspromisesmkdirpath-options

hereje: > Shouldn't this be using the promise version of `mkdir`? https://nodejs.org/api/fs.html#fspromisesmkdirpath-options

I agree.

What should we do when eslint shows the following error?

The 'fs/promises' is not supported until Node.js 14.0.0. The configured version range is '>=8.0.0'.eslint[n/no-unsupported-features/node-builtins](https://github.com/weiran-zsd/eslint-plugin-node/blob/HEAD/docs/rules/no-unsupported-features/node-builtins.md)


Should we update eslint or just silent the validation error?


The project currently uses 

```
"eslint": "^8.38.0",
"eslint-config-prettier": "^8.8.0",
"eslint-plugin-import": "^2.27.5",
"eslint-plugin-n": "^15.7.0",
"eslint-plugin-prettier": "^4.2.1",
```

And the `.nvmrc` indicates to use node `14`

kitop: Good question!
We're only testing on latest and as you said the `.nvmrc` indicates 14. We should probably also make it so that the `engines` property in `package.json` indicates >=14 as well, and that'd make it so that eslint picks that version as the minimum req for that rule, according to [the docs](https://github.com/eslint-community/eslint-plugin-n/blob/HEAD/docs/rules/no-unsupported-features/node-builtins.md#version).

hereje: > Good question! We're only testing on latest and as you said the `.nvmrc` indicates 14. We should probably also make it so that the `engines` property in `package.json` indicates >=14 as well, and that'd make it so that eslint picks that version as the minimum req for that rule, according to [the docs](https://github.com/eslint-community/eslint-plugin-n/blob/HEAD/docs/rules/no-unsupported-features/node-builtins.md#version).

So, I propose the following changes:

- rename PR as **refactor: replace promisify with fs/promises version convert.js**
- node version in `.nvmrc` as `21.5.0`
-  node engine >=14 in  `package.json`  `"engines": { "node": ">=14" }`
- and later create a PR in order to migrate `promisify` with the promise version of mkdir, writeFile throughout the project

What do you think?

kitop: I wouldn't touch the `.nvmrc` file in this change, so we can keep it to a minimum. If we're saying the minimum supported version is 14, it's not bad to have that in the local setup. I'd keep it as-is for now, and we can look at the local version changes at some other time (maybe with testing or some other JS setup).

Other than that, I think the plan makes sense!

src/convert.js
```diff
@@ -15,10 +11,10 @@ const JSON_OUTPUT = `${OUTPUT_DIR}/swagger.json`
...
```
kitop, line 14: The promise version doesn't take a third argument for the callback ([docs](https://nodejs.org/api/fs.html#fspromisesmkdirpath-options))
```suggestion
    mkdir(OUTPUT_DIR, { recursive: true }),
```

kitop (commented):
One small comment. And then I'm curious if there's a way to make the package.json have less changes. Since there's a lot because it's changing from V1 to V3 - would it be possible to have it keep the old format? It'd be nicer to change the format in a specific PR.

hereje:
> One small comment. And then I'm curious if there's a way to make the package.json have less changes. Since there's a lot because it's changing from V1 to V3 - would it be possible to have it keep the old format? It'd be nicer to change the format in a specific PR.

Sorry, My mistake.


I will do it through:

```
npm install --lockfile-version 1
```

kitop:
@hereje thanks for the update! Unfortunately, seems like there's still a lot of additions to the package-lock.json, like adding webpack, source-map and others that were not there before.

I've tried the same change in the package.json locally (remove `mk-dir` and add the engines definition) and here's the diff I get: https://gist.github.com/kitop/9b53d60e4b2650bc0a03f6036ac76477

Any reason why the other dependencies where added? I'd like to keep the package-lock.json as clean as possible.

hereje:
> @hereje thanks for the update! Unfortunately, seems like there's still a lot of additions to the package-lock.json, like adding webpack, source-map and others that were not there before.
> 
> I've tried the same change in the package.json locally (remove `mk-dir` and add the engines definition) and here's the diff I get: https://gist.github.com/kitop/9b53d60e4b2650bc0a03f6036ac76477
> 
> Any reason why the other dependencies where added? I'd like to keep the package-lock.json as clean as possible.

Yeah, it seems I poluted the file when I made a dependency install.

I have fixed the issue.

kitop (approved):

PR #528, opened 2024-01-29
feat: Replace `Controller` role with `Billing Admin` role

Ayushi3333:
We recently renamed all Controllers to Billing Admins in [PR](https://github.com/netlify/bitballoon/pull/17481). This needs to be reflected in the api docs.


mraerino (commented):
this seems like a breaking change? should we have a backwards compatible API that allows `Controller` as an alias for `Billing Admin`?

also i think you need to run `go generate ./...` and commit the generated files

Ayushi3333:
> this seems like a breaking change? should we have a backwards compatible API that allows `Controller` as an alias for `Billing Admin`?
> 
> also i think you need to run `go generate ./...` and commit the generated files

Can you elaborate how this is a breaking change? We no longer accept `Controller` as a role in the API requests. This change is ensuring `Controller` is no longer used because it won't work.

Thanks for the second suggestion, I will try to generate the files.

mraerino:
> Can you elaborate how this is a breaking change? We no longer accept `Controller` as a role in the API requests. This change is ensuring `Controller` is no longer used because it won't work. If you are worried about our UI, `netlify-react-ui` is sending both roles in the params to support the transition.

yeah, exactly what i meant. we made a breaking change to our API since existing code from customers that might have used `Controller` now fails.
i'm bringing this up here because everything that we document in `open-api` is supposed to be fairly stable, since once it's documented publicly, customers will start relying it. if we hadn't documented it, the breaking change wouldn't be too significant since we don't need to provide stability when we expect a method to only be used in our UI.

we have this distinction deliberately. there is an [internal open api spec](https://github.com/netlify/bitballoon-openapi) that contains all methods that our API offers, not just the ones we want to provide to customers

Ayushi3333:
> > Can you elaborate how this is a breaking change? We no longer accept `Controller` as a role in the API requests. This change is ensuring `Controller` is no longer used because it won't work. If you are worried about our UI, `netlify-react-ui` is sending both roles in the params to support the transition.
> 
> yeah, exactly what i meant. we made a breaking change to our API since existing code from customers that might have used `Controller` now fails. i'm bringing this up here because everything that we document in `open-api` is supposed to be fairly stable, since once it's documented publicly, customers will start relying on it. if we hadn't documented it, the breaking change wouldn't be too significant since we don't need to provide stability when we expect a method to only be used in our UI.
> 
> we have this distinction deliberately. there is an [internal open api spec](https://github.com/netlify/bitballoon-openapi) that contains all methods that our API offers, not just the ones we want to provide to customers

You are absolutely right that this would be a problem if this was being actively used. There are 85 non-netlify Billing Admins all time and I have a strong suspicion they were not created through the API but rather through the UI. 

I do not think this role is being created by the users programatically. To substantiate my claim, I tried looking in Humio but I cannot find a way to differentiate "user-generated" requests vs our UI. I can find some data to confirm once I figure out how to obtain it.

mraerino (approved):

PR #557, opened 2025-01-17
fix: update `getCurrentUser` response type from array to single `User` object

CalebBarnes:
`getCurrentUser` response schema was incorrectly set as an array but should just be a single user object.


khendrikse (approved):
Good eye

PR #650, opened 2026-06-30
feat: new edge function upload handling

pieh:
Relies on https://github.com/netlify/bitballoon/pull/22198 (merged) + https://github.com/netlify/netlify-server/pull/4267 (also merged)

Part of https://linear.app/netlify/issue/RUN-2937/implement-end-to-end-env-vars-plumbing-for-functions


go/porcelain/deploy.go
```diff
@@ -1077,6 +1113,67 @@ func zipFunctionFile(filePath string, i os.FileInfo, runtime string, tmpDir *laz
 	}, nil
 }
 
+// bundleEdgeFunctions reads the edge-bundler manifest from edgeFunctionsDir and turns each bundle it
+// lists into an uploadable FileBundle. The deploy declares these as its edge_functions map
+// ({format => code_sha}); the server replies with the subset (required_edge_functions) not already
```
pieh, line 1118: Note - in here we send everything in edge-bundler manifest - including `eszip` bundles. Bitballoon intentionally only cares about `tar` ( https://github.com/netlify/bitballoon/blob/ad4d323eb9f9fabf68c5518e4ce701ca9faa964e/app/models/deploy.rb#L1114-L1119 ) 

Goal of this was to make clients to not to make such decisions and instead have our backed alone decide about what goes through which upload channel.

This way it's much easier for us to control things. It's less relevant for `porcelain` client which we can rollout with buildbot anytime we want. But the same will be replicated for CLI where impact of this decision is bigger, because we can't fully control over which version is actually used and this should allow us to just eventually turn off tar uploads going through static files uploads channel

swagger.yml
```diff
@@ -1665,6 +1665,33 @@ paths:
...
+          schema:
+            type: string
+            format: binary
+          required: true
+        - $ref: '#/parameters/retryCount'
+      responses:
+        '200':
+          description: OK
+        default:
+          $ref: '#/responses/error'
```
pieh, line 1694: I think it's worth doing some extra check here, but here's claude's summary of "correctness" of this:
```
The endpoint definition is correct — it will interoperate properly with the server. I traced every part of the swagger spec against the handler. Here's the validation:

┌─────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────┬───────────────────────┐
│           Swagger (swagger.yml:1668-1694)           │                     Server (request_edge_function_upload.go / server.go)                      │         Match         │
├─────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────┤
│ Path /deploys/{deploy_id}/edge_functions/{code_sha} │ Regex ^/api/v1/deploys/(?P<deploy_id>[a-f0-9]+)/edge_functions/(?P<code_sha>[a-f0-9]+)$ (:19) │ ✅                    │
├─────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────┤
│ Method put                                          │ Test uses http.MethodPut → 200 (:703-705)                                                     │ ✅ (works)            │
├─────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────┤
│ consumes: application/octet-stream                  │ Body streamed raw to UploadEdgeFunctionBundle (:82)                                           │ ✅                    │
├─────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────┤
│ file_body binary, required                          │ Reads req body                                                                                │ ✅                    │
├─────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────┤
│ retryCount param                                    │ not read by handler                                                                           │ ⚠️ ignored (harmless) │
├─────────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────┼───────────────────────┤
│ Response 200, no schema (empty body)                │ Returns nil → default 200, empty body (:93); test asserts StatusOK                            │ ✅                    │
└─────────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────┴───────────────────────┘

Two minor, non-blocking discrepancies worth knowing

1. Method isn't enforced server-side. The dispatch at server.go:657 matches on URL path only (isFOEdgeFunctionUploadPath(req.URL.Path)), with no method check — unlike the swagger which declares put. So the server would actually accept any verb on that path. This isn't a bug for you: the generated client always issues PUT, so declaring put in swagger is the correct, tighter contract. Just don't expect a GET/POST to that path to be rejected with a 405.

2. retryCount is sent but ignored. Your client sends X-Nf-Retry-Count (params.WithXNfRetryCount), and the swagger declares the retryCount param — but handleFOEdgeFunctionUpload never reads it. So retries work (the client re-PUTs), but the retry-count header has no server-side effect on this endpoint (no metrics/dedup keyed on it). Contrast with the function-upload path, which does thread it through. Not a correctness issue — worth a mental note if you ever expect the server to behave differently on retries.
```

I did use `- $ref: '#/parameters/retryCount'` to mimic function uploads - `netlify-server` similarly doesn't do anything with it - so it's a bit useless, but also seems harmless to have it ... at least it will be consistent with already existing functions definition

lemusthelroy (approved):

